PR #6, opened 2015-07-21
Update readme for using library on HTML elements

cdl:
- Update README to properly denote how to use the library with inline HTML elements
- add `data-` to HTML example attributes
- add `has-tooltip` class to each element with a tooltip, as well as include instructions denoting as such
- taken from https://github.com/sir-dunxalot/ember-tooltips/blob/master/addon/mixins/components/tooltips.js#L127-L184

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,15 +110,15 @@ camelCased Options can still be passed to the component but they are not prefixe
 
 ### Using on HTML elements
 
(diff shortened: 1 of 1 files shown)


sir-dunxalot:
Looks great, thank you. I'll push the changes up to NPM with the next minor or patch release.

Issue #5, opened 2015-07-21
Tooltip text doesn't change when tooltipContent changes


cdl:
I have the following element inside a component:

```hbs
<a class="btn btn--grade has-tooltip"
    {{action "grade"}}
    data-tooltip-content={{gradeTooltipText}}>
  {{!-- button content... --}}
</a>
```

The component, on `didInsertElement`, calls `this.renderChildTooltips();`, which works great. However, **when the `data-tooltip-content` attribute changes, the content of the tooltip itself doesn't change**.

Subsequent calls to `this.renderChildTooltips()` doesn't re-render the existing tooltip - it just creates another tooltip with the new content, the old one still existing.

Is there a method I can call to destroy and re-render (or update) all tooltips? Or is this just a case that isn't covered yet?

sir-dunxalot:
I think this is very valid, especially with the nature of promises.

The [tooltip API doc are available here](https://github.com/darsain/tooltip/wiki/Tooltip). There is a `Tooltip.detach()` method, which might suit our needs here. However, better still is the `Tooltip.content()` method, which changes the tooltip content - so it's definitely possible.

One blocker here is that whilst the tooltip is referenced as `tooltip` on components and views, the tooltips rendered using `renderChildTooltips` are not saved for later reference. This shouldn't be too hard to do. This would probably require each tooltip to have a unique ID so you can change to correct one.

Perhaps `renderChildTooltips()` and associated functionality could even be moved into its own mixin.

I'd be more than happy to accept a PR for this as it will probably take me a couple of weeks to get around to it.

jamesarosen:
`renderTooltip` already [sets](https://github.com/sir-dunxalot/ember-tooltips/blob/3faafa5105d3f46cc58480aadfdc845081f0d772/addon/mixins/components/tooltips.js#L149) `tooltip` to the `Tooltip` instance.

Thus, the simplest thing that could _possibly_ work would be
```js
_updateTooltipContent: Ember.observes('tooltipContent', function() {
  const tooltip = this.get('tooltip');
  if (tooltip == null) { return; }

  tooltip.content(this.get('tooltipContent'));
})
```

I don't like that because it uses an Observer. It also might not work if the component that has `tooltipContent` differs from the one that has `tooltip`, which might happen with `{{#tooltip-on-parent}}`; I'm not sure

A more "correct" approach would be
```js
didReceiveAttrs({ newAttrs }) {
  const tooltip = this.get('tooltip');
[...]

PR #9, opened 2015-07-23
Feature/manual trigger

davidgovea:
Hello sir-dunxalot - great plugin you got here! So many flexible usage options - the `{{tooltip-on-parent}}` with HTMLBars block for content is a great touch.

We have a use-case where we'd like to show more "permanent" tooltips on certain UI elements. I've extended your plugin to allow for manual triggering. Tried to follow your style & make clean, incremental commits.

Some notes:

* I decided to extend the `event` option after I verified that `darsain/tooltips` is not using that property. I could have made manual-mode a separate option, but since `event` is specific to this plugin, I figured `event: 'manual'` would be acceptable. Let me know if you'd prefer another approach.
* Specifying `event:'manual'` and binding `open` works for both on-component and `{{tooltip-on-parent}}` use-cases, but *not* for HTML data-attrs. Seems architecturally related to issue #5, so I didn't dive in there.

Let me know your thoughts! If I'm on a good track, I can definitely provide more tests for these features.

### DIFF
--- a/addon/mixins/components/tooltips.js
+++ b/addon/mixins/components/tooltips.js
@@ -41,6 +41,7 @@ export default Ember.Mixin.create({
   tooltipPlace: 'top',
   tooltipSpacing: 10,
   tooltipTypeClass: null,
+  tooltipOpen: null, // for manual-mode triggering
 
   /**
   Removes a tooltip from the DOM if the element it is attached
--- a/addon/mixins/components/tooltips.js
+++ b/addon/mixins/components/tooltips.js
@@ -58,6 +58,9 @@ export default Ember.Mixin.create({
       tooltip.effect(null); // Remove animation
       tooltip.detach();
     }
+
+    /* Remove observer, even if it was never added */
+    this.removeObserver('tooltipOpen', this, this.tooltipOpenDidChange);
   }),
 
   /**
--- a/addon/mixins/components/tooltips.js
+++ b/addon/mixins/components/tooltips.js
@@ -144,6 +144,23 @@ export default Ember.Mixin.create({
   }),
 
   /**
+  Opens / closes tooltip based on value of 'tooltipOpen'.
+  Only used when event is 'manual'.
+
+  @method tooltipOpenDidChange
+  */
+
+  tooltipOpenDidChange: function() {
+    const tooltip = this.get('tooltip');
+
+    if (this.get('tooltipOpen')) {
+      tooltip.show();
+    } else {
+      tooltip.hide();
+    }
+  },
+
+  /**
   Call this method on any view to attach tooltips to all elements in its
   template that have a `.tooltip` class. Tooltip options are set using
   data attributes.
--- a/addon/mixins/components/tooltips.js
+++ b/addon/mixins/components/tooltips.js
@@ -37,7 +37,7 @@ export default Ember.Mixin.create({
   tooltipAuto: true,
   tooltipContent: null,
   tooltipEffectClass: 'slide', // fade, grow, slide, null
-  tooltipEvent: 'hover',
+  tooltipEvent: 'hover', // DOM event or "manual"
   tooltipPlace: 'top',
   tooltipSpacing: 10,
   tooltipTypeClass: null,
--- a/addon/mixins/components/tooltips.js
+++ b/addon/mixins/components/tooltips.js
@@ -29,7 +29,8 @@ export default Ember.Mixin.create({
     'event',
     'place',
     'spacing',
-    'typeClass'
+    'typeClass',
+    'open'
   ],
 
   /* Tooltip options - see http://darsa.in/tooltip/ */
@@ -144,6 +145,12 @@ export default Ember.Mixin.create({
 
     /* Bind observer if manual-triggering mode */
     if (tooltipOptions.event === 'manual') {
+      if (componentWasPassed) {
+        // Keep track of child tooltip component
+        this.set('tooltipChildComponent', component);
+        // Turn 'tooltipOpen' into a computed property, reading from child tooltip component's 'open' option
+        Ember.defineProperty(this, 'tooltipOpen', Ember.computed.reads('tooltipChildComponent.open'));
+      }
       Ember.addObserver(this, 'tooltipOpen', this, this.tooltipOpenDidChange);
       this.tooltipOpenDidChange();
     }
--- a/addon/mixins/components/tooltips.js
+++ b/addon/mixins/components/tooltips.js
@@ -30,7 +30,7 @@ export default Ember.Mixin.create({
     'place',
     'spacing',
     'typeClass',
-    'open'
+    'visibility'
   ],
 
   /* Tooltip options - see http://darsa.in/tooltip/ */
@@ -42,7 +42,7 @@ export default Ember.Mixin.create({
   tooltipPlace: 'top',
   tooltipSpacing: 10,
   tooltipTypeClass: null,
-  tooltipOpen: null, // for manual-mode triggering
+  tooltipVisibility: null, // for manual-mode triggering
 
   /**
   Removes a tooltip from the DOM if the element it is attached
@@ -60,7 +60,7 @@ export default Ember.Mixin.create({
     }
 
     /* Remove observer, even if it was never added */
-    this.removeObserver('tooltipOpen', this, this.tooltipOpenDidChange);
+    this.removeObserver('tooltipVisibility', this, this.tooltipVisibilityDidChange);
   }),
 
   /**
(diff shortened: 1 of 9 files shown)


sir-dunxalot:
Hey David, this is a great feature add. Thanks for putting the work into it. I know I will use it as soon as it's merged!

So I'm clear here, am I correct in saying `tooltipOpen` is a boolean property that only takes effect when the existing `tooltipEvent` property is set to `'manual'`?

I've give the code a proper look through in the morning. Overall looks great. I have a couple of thoughts thoughts:

- Could we run into some issues with tooltipOpen being used with other events? What if `tooltipOpen` could override the event regardless of what `tooltipEvent` is equal to?
- We can probably avoid observers all together by using the new component life cycle hooks. This would make for cleaner code but would break pre-1.13 Ember apps. We could add a check to see if the app is running `1.13`+ and not add the observer in that case. It would be really easy to support 1.13+ by wrapping `tooltipOpenDidChange` in an `Ember.on('didReceiveAttrs'` or `Ember.on('didUpdateAttrs'`. However, this all sounds like a lot of work for small benefit. Maybe that's a 'TODO'?
- If we keep the observers we should remove them using `this.removeObserver` when the component is destroyed. I added a hook called `destroyTooltip` yesterday for this kind of work.
- We should add an acceptance test for this. The way the current testing is setup is a route for each main feature. Thus, we could add a route in the dummy app just for static tooltips.

sir-dunxalot:
Also, I'm happy to help with the testing. Just let me know.

davidgovea:
Yo dunx, thanks for the feedback!

Made sure to clean up my observer mess, and added some tests. Not sure if I'm properly using the test helper's `content` and `targetContent` properties [here](https://github.com/sir-dunxalot/ember-tooltips/commit/09df12d05a34ff1013063c3256679b9ccff5fd7f#diff-f09b76b8498f214806ec0e966ce598e8R28) in the manual+component test.

As for your concern of `tooltipOpen` conflicting with the normal event mode, it shouldn't be a problem because the observer is [only added when `event === 'manual'`](https://github.com/runtrizapps/ember-tooltips/blob/feature/manual-trigger/addon/mixins/components/tooltips.js#L150). I'd be down to add a test for that though, when I get another minute. Also, I'm assuming that the event *doesn't change* - I think that's reasonable.

As far as the 1.13+ component hooks - good call there! Most all of my projects are up to 1.13 at this point & I've been itching to use those. Would you prefer a runtime check (`Ember.VERSION`) or a cli build-time check that manipulates broccoli tree depending on version?

tip-tests-compress

sir-dunxalot:
Awesome. I'm not sure about the version checking - I've never done that before. What do you think?

davidgovea:
Slapping a conditional on `Ember.VERSION` sounds easier ("good enough"?), but there is something kinda elegant about the cli tooling approach. I've never done ember version checking either, though.

sir-dunxalot:
Given it's (hopefully) a short lived feature and is only adding one extra method to the class (instead of defining a different class altogether) I think we'll be ok with the `Ember.VERSION`. That runs in the browser, correct?

addon/mixins/components/tooltips.js
```diff
@@ -140,9 +145,38 @@ export default Ember.Mixin.create({
...
+
+    /* Bind observer if manual-triggering mode */
+    if (tooltipOptions.event === 'manual') {
+      if (componentWasPassed) {
+        // Keep track of child tooltip component
+        this.set('tooltipChildComponent', component);
+        // Turn 'tooltipOpen' into a computed property, reading from child tooltip component's 'open' option
+        Ember.defineProperty(this, 'tooltipOpen', Ember.computed.reads('tooltipChildComponent.open'));
+      }
+      Ember.addObserver(this, 'tooltipOpen', this, this.tooltipOpenDidChange);
```
sir-dunxalot: Maybe `this.addObserver` for code consistency with the `this.removeObserver` call?

tests/acceptance/tooltip-manual-trigger-test.js
```diff
@@ -0,0 +1,35 @@
...
+});
+
+test('Rendering tooltips with event="manual" and provided "tooltipOpen" boolean', function(assert) {
+  visit('/tooltip-manual-trigger');
+
+  assert.expect(13);
+
+  assertTooltipProperties(assert, 'manually-trigger', {
+    content: 'This is a manually triggered tooltip',
+    usingComponent: false,
```
sir-dunxalot: You don't have to pass `usingComponent` unless you're overriding the default.

tests/acceptance/tooltip-manual-trigger-test.js
```diff
@@ -0,0 +1,35 @@
...
+    event: 'manual'
+  });
+
+  assertTooltipProperties(assert, 'manually-trigger-component', {
+    targetContent: 'Manual tooltip as component',
```
sir-dunxalot: To your question, this use is perfect. :+1:

tests/helpers/async/assert-tooltip-properties.js
```diff
@@ -34,6 +34,8 @@ export default Ember.Test.registerAsyncHelper('assertTooltipProperties',
       click(selectorFor(name));
     } else if (expectedEvent === 'hover') {
       mouseOver(name);
+    } else if (expectedEvent) {
```
sir-dunxalot: Should this be `} else if (expectedEvent === 'manual') {`?

tests/helpers/async/assert-tooltip-properties.js
```diff
@@ -125,8 +127,11 @@ export default Ember.Test.registerAsyncHelper('assertTooltipProperties',
       click(selectorFor(name));
     } else if (expectedEvent === 'hover') {
       mouseOut(name);
+    } else if (expectedEvent) {
```
sir-dunxalot: As above, should this be `} else if (expectedEvent === 'manual') {`?

sir-dunxalot:
Just had a look through the code - everything is looking really slick. Great work! This can also solve issues like #10. I'll leave it up to you as to whether you want to add the observer-less code too.

If you do that I will add Ember `1.12` to the Ember Try test suite set.

But either way let me know when you're ready to merge this in.

davidgovea:
Cleaned things up per your comments - I'd say this is ready to go! The 1.13+ stuff can be a future enhancement. Working on it already.

I started a [branch](https://github.com/runtrizapps/ember-tooltips/tree/feature/manual-trigger-newhooks) with the new component hooks (blowing away <1.12 support for now, just to explore), but there are things that should be worked out.

Mainly, I'd like to avoid keeping track of the child `{{tooltip-on-parent}}` component, and instead use actions. You could say that the dynamic computed prop that reads child open state is "data UP" rather than data-down, actions-up. If observers must be used for <1.12, then the `tooltip-on-parent` should implement its own as well, rather than the computed-prop fanciness.

One cool part about using actions for the manual-triggering would be built-in input action triggering:

**templates/components/input-with-tooltip.hbs**
```hbs
{{#tooltip-on-parent event='manual'}}
  Manual trigger
{{/tooltip-on-parent}}
{{input type="text" focus-in="showTooltip" focus-out="hideTooltip" value=value}}
```

davidgovea:
Soo... was having lots of fun with the new component hooks and I got pretty far. 
You can compare the branch to this one [here](https://github.com/runtrizapps/ember-tooltips/compare/feature/manual-trigger...runtrizapps:feature/manual-trigger-newhooks?diff=split&name=feature%2Fmanual-trigger-newhooks)

Most notable change is the use of `showTooltip` / `hideTooltip` actions, and not keeping track of the child component in the parent.
Not sure how I feel about those action names, though - since we're reopening a lot of things by default. Perhaps keeping a "tooltip" prefix like the other properties would be better: `tooltipShow` / `tooltipHide`

All of the tests work on both Ember 1.12 and 1.13 (added to the ember-try config), except for this superficial failure in 1.12 (rendering engine swapped order of `href` and `class`):
```
not ok 7 PhantomJS 2.0 - Acceptance | tooltip as component: Rendering tooltips set using {{tooltip-on-parent}}
    ---
        actual: >
            <a class="ember-view" href="/">Using HTMLBars</a> in content
        expected: >
            <a href="/" class="ember-view">Using HTMLBars</a> in content
        message: >
            The HTML content of the tooltip should be correct
        Log: >
    ...
```

I think having the action API will be another good addition to the flexibility of this library. It might be convenient for implementing #10, and also convenient for component authors to interact with a tooltip. Note that the action's **don't** verify that you're in manual mode, by design. We may want to say, auto-close after a click-open by sending the action.

Let me know what you think & I can clean up some things - the version detection is ugly, for one..

sir-dunxalot:
Very cool stuff! I'm stuck on another project today, which is why I haven't released yet. Do you want to go ahead and release this PR or would you rather wait until the 1.13 compat is set?

I have a look through the other branch. How would one go about implementing the actions?

davidgovea:
The new stuff is a good deal more code for no huge functional improvement. 
It's also my first time working with the new hooks (think I'm doing it right..). So maybe we don't rush it, merge what you've reviewed here, and then I can PR the enhancement for better discussion.

sir-dunxalot:
Ok, sounds good. I'll do a final look through tomorrow and then we'll put this out as the next release.

sir-dunxalot:
Had an idea this evening: what do you think about renaming `tooltipOpen` to `tooltipVisibility`?

davidgovea:
I like it! ..especially if actions are coming into the mix.. much less ambiguous
even `tooltipVisible` (I'm thinking of Ember `isVisible`.. is that even a thing any more?)

sir-dunxalot:
Then maybe `tooltipIsVisible`? Haha. I'm cool with either `tooltipVisiblity` or `tooltipIsVisible`. Once that's done let's get this baby merged!

davidgovea:
:checkered_flag:

sir-dunxalot:
Great work! Released as `v0.3.0`.

PR #20, opened 2015-09-10
Observing changes to tooltipContent property.

lukes:
Hi, 

Not sure if you want/need this PR. It adds a specific functionality; observing changes to the `tooltipContent` property and updating the content of the tooltip. 

I noticed #5 had a similar issue, but this fix doesn't address their problem.

### DIFF
--- a/addon/mixins/components/tooltips.js
+++ b/addon/mixins/components/tooltips.js
@@ -114,8 +114,8 @@ export default Ember.Mixin.create({
     let content = this.get('tooltipContent');
     let tooltip, tooltipOptions;
 
-    var keys = Object.keys(this);
-    var hasTooltipContentProperty = (Ember.$.inArray('tooltipContent', keys) !== -1);
+    const keys = Object.keys(this);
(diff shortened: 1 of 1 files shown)


sir-dunxalot:
Awesome, thank you. And +1 for issuing to the `2.0-release` branch too. I'll have another look through in the morning but this looks like a great addition to the addon.

addon/mixins/components/tooltips.js
```diff
@@ -113,6 +114,9 @@ export default Ember.Mixin.create({
     let content = this.get('tooltipContent');
     let tooltip, tooltipOptions;
 
+    var keys = Object.keys(this);
```
sir-dunxalot: Can you use `const` instead of var for consistency?

sir-dunxalot:
One small comment. Otherwise this looks good!

lukes:
Thanks, glad it's useful to you! I've made those changes. Thanks for maintaining this library!

sir-dunxalot:
Thank you and my pleasure - I'll try and get the 2.0 compatible version released as soon as possible.

PR #23, opened 2015-10-06
Fix an issue where component teardown breaks IE

maxhungry:
document.contains() is not available in IE, it returns an error message
"Object doesn't support property or method 'contains'", (tested on both IE10 &
IE11). Use document.body.contains() instead.


sir-dunxalot:
Totally my fault - sorry for not thinking of IE.

Thanks for testing it. I'll pull down and release right now.

sir-dunxalot:
Released as `0.5.2` on NPM and Github. Thanks a lot for the PR!

PR #33, opened 2015-12-26
Event refactoring

sir-dunxalot:
Fixes some issues, improves documentation, and adds `hideOn` and `showOn`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,38 +11,28 @@ Powered by <a href="http://darsa.in/tooltip/" target="_blank">darsain/tooltip</a
 ember install ember-tooltips
 ```
 
-## Usage
+## Documentation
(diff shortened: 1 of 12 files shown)


eduardbaitinger:
Commit b248595 adds tooltip content also to the title attribute and enables the browser tooltip.
This leads to duplicated tooltips and can be very ugly when a tooltip contains html-markup, which is also displayed in the native browser tooltip.

sir-dunxalot:
Agreed. This was a regression. I will add an issue not we keep track of that.

sir-dunxalot:
See #38

eduardbaitinger:
Ok, thank you for the quick response.

PR #34, opened 2016-01-11
Allow tooltip HTML titles to have an empty string

mattboldt:
When using HTML in tooltips, it doesn't look good for that to appear in the standard HTML tag `title` attr (when hovered over the element). This allows you to set the `title` to an empty string. If it's not an empty string, it's undefined and goes back to using `content.toString()`


webark:
Was this, or something like this, going to get merged in?

Issue #24, opened 2015-10-07
Tooltip reappears after route transition


jamesarosen:
If I add a tooltip to a link:
```hbs
{{link-to "Somewhere" "some.where" tooltipContent="Go Somehwere"}}
```
then the tooltip shows up when I hover over the link (yay!). When I click the link, the tooltip disappears (yay!). But when I then move the mouse, it reappears (boo!).

I've set a debugger on `Tooltip.prototype.show`. It turns out that the `mouseover` event is getting fired on the `<a>` when it no longer is in the DOM. I think the simple solution is to guard in either `Tooltip.prototype.show` (an upstream library unfortunately) or in `renderTooltip`:

```js
var willShow = tooltip.hidden && jQuery.contains(document, tooltip.element);
```

sir-dunxalot:
Thanks for digging into it. Which version of this addon are you using? `0.5.2` should contain a fix for this issue.

jamesarosen:
We were on 0.4.0. I see 53d3cfb34a5b5012656a62a514d573c95bd9d48e. I'm trying 0.5.2 now :)

sir-dunxalot:
Ok, great. Let me know if the issue still persists.

jrjohnson:
Is it possible to release a fix on the 0.4.x tree that will work with Ember 1.13?  We're still cleaning out deprecations so we can get into 2.0.  Or am I reading the README wrong and 0.5.x will work with ember pre 2.0?

sir-dunxalot:
I removed the compatibility table because 0.5.x should be backwards compatible. What deprecation warnings are you seeing?

jrjohnson:
All deprecations are in my app.  I hadn't even tried 0.5.x - didn't think it was for me.  I will give it a shot, thanks.

sir-dunxalot:
Ok. If there are any problems just let me know.

PR #36, opened 2016-01-17
Bug fix for tooltips on link-to components not disappearing.

pauc:


sir-dunxalot:
Pau, you're a lifesaver! This bug has been an issue for a while. I will pull down, test it out, and merge ASAP.

sir-dunxalot:
Fixes #24 (again).

pauc:
I added some test but they are not really useful, because I couldn't get a fail before apply the fix, so it was a green to green...

sir-dunxalot:
Ok. I'll see if I can cause a failing test on master branch before merging this.

pauc:
I think part of the problem is caused by this lines: https://github.com/sir-dunxalot/ember-tooltips/blob/0a8842e957f1361c1e64c707d084e588a0f4b8b6/addon/utils/render-tooltip.js#L194 and the 'debounce' on line 39. The 'focusin' event is fired alongside with the 'click' event, and this line https://github.com/sir-dunxalot/ember-tooltips/blob/0a8842e957f1361c1e64c707d084e588a0f4b8b6/addon/utils/render-tooltip.js#L56 gets executed when the component with the tooltip is on the "isDestroying" state and the transition to the new route already took place.

I created a brach with some debug statements to try to see what happens: https://github.com/pauc/ember-tooltips/tree/debug.

I really can't figure out how to reproduce the bug in acceptance tests...

Hey, and thanks for the addon :-) !!

sir-dunxalot:
I think I have a failing acceptance test. I also have a reproducible bug for manual testing. I will go ahead and merge this to see what happens.

sir-dunxalot:
The failing test didn't pass when this was merged but the reproducible bug has been fixed. We at least have *some* test coverage against transitions. My worry is the bug is rooted in something not reproducible calling `click()` in the JS but, instead, must be triggered by a user clicking in the browser.

However, I've done a lot of digging without much progress and this PR fixes the issue. I will release shortly because this bug is affecting many people. I will also create an issue for not having a failing test for the bug fixed by this PR.

sir-dunxalot:
cc @cmcn

iezer:
Does using `tooltip-for-component` inside a `link-to` work? I can't get it to work and I'm using ember-tooltips 2.5.0 and ember 2.7.3. Also is the `tooltipContent` property deprecated?

PR #41, opened 2016-01-27
If event is manual or none, bypass all other show/hide event triggers.

### LINKED ISSUE #40: Tooltip made visible on focus even when event set to "manual"
Hey, was a little perplexed as to why clicking on my target element was causing my tooltip visibility to toggle, given that I had set the visibility of the tooltip to manual.

After a combing through the code (very well written btw, great documentation) I found that ember-tooltips toggles visibility on focus for any event not set to "focus"; a little counter-intuitive, but the comment above made it clear that this was specifically for accessibility. 

https://github.com/sir-dunxalot/ember-tooltips/blob/43842d88435151fbdd68b5164076b76263e6bb29/addon/utils/render-tooltip.js#L196

While I think the intent is clear and fair, the primary use case for "manual" is to have explicit control for when a tooltip needs to be present and I think this undermines the use case. 

Might I suggest we change that line to:
```js
if (event !== 'focus' && event !=='manual' && event!=='none') {
```

webark:
Even ones for accessibility. Fixes #40

### DIFF
--- a/addon/utils/render-tooltip.js
+++ b/addon/utils/render-tooltip.js
@@ -188,29 +188,29 @@ export default function renderTooltip(domElement, options, context) {
         });
       }
     }
-  }
 
-  /* Hide and show the tooltip on focus and escape
-  for acessibility */
+    /* Hide and show the tooltip on focus and escape
+    for acessibility */
 
-  if (event !== 'focus') {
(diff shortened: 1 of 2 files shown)


sir-dunxalot:
Good stuff. I am planning to merge and release this sometime this week.

PR #44, opened 2016-02-08
Remove the deprecation messages

ksnyde:
By switching to `init()` (within a "afterRender" schedule) we can avoid the ugly deprecations that often creep into `didInsertElement`.

### DIFF
--- a/addon/components/tooltip-on-parent.js
+++ b/addon/components/tooltip-on-parent.js
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 import layout from '../templates/components/tooltip-on-parent';
 
-const { computed, warn, run, get } = Ember;
+const { computed, run, get } = Ember;
 
 export default Ember.Component.extend({
   attributeBindings: ['style'],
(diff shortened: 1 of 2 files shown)


ksnyde:
This now just warns people who have no tagName without providing a fix. It also fixes the deprecation warning mentioned earlier.

sir-dunxalot:
Any idea why the test suite is failing with this branch?

addon/components/tooltip-on-parent.js
```diff
@@ -11,16 +11,22 @@ export default Ember.Component.extend({
...
 
-    if (parentView.renderTooltip) {
-      parentView.renderTooltip(this);
-    } else {
-      warn('No renderTooltip method found on the parent view of the {{tooltip-on-parent}} component');
-    }
+    run.schedule('afterRender', () => {
+      const parentView = this.get('parentView');
+      const componentNotInDOM = get(parentView, 'tagName') ? false : true;
+      if(componentNotInDOM) {
```
sir-dunxalot: Why not just use:

```js
if (!parentView.get('tagName')) {
  console.warn('etc');
} else if...
```

sir-dunxalot: My point being that returning boolean values from a ternary operation seems a little redundant.

sir-dunxalot:
This looks good to me once we resolve the test suite issues.

addon/components/tooltip-on-parent.js
ksnyde: Yeah it was just to be more descriptive about why we care about `tagName` but I'm happy to do it your way. I'll change that now and fix unit testing error.

BTW, sorry for the number of typos and UT failures, I'm in running fast mode and most my edits came from within Github :p

sir-dunxalot: Hah, totally cool.

sir-dunxalot:
I'm going to merge this and will fix the test suite locally.

Issue #47, opened 2016-02-26
Blocker bug - plugin explodes if you already have a tooltip property anywhere in your app or any of your apps components


coladarci:
This plugin adds a `destroyTooltip: on('willDestroyElement', function () {` method that is in the mixin that gets injected everywhere.

Inside there it says:

```
var tooltip = this.get('tooltip');

if (tooltip) {
  .... DO SOME CRAY STUFF THAT ONLY MY TOOLTIP RESPONDS TO
}
```

So, if you have a tooltip property, or worse yet, you are using someone else's add-on who has a tooltip property, your app will come to a screeching halt when you navigate away from that view.

The fix should be that this plugin uses a more specific name for that property that is less likely to conflict AND checks to see if the `tooltip` it gets back is, in fact, the type of tooltip you are expecting.

sir-dunxalot:
I think the second part of that (checking the type/constructor of `tooltip`) is reasonable. The first part, about renaming the `tooltip` property seems like more of an odd edge case. My reasoning being that this is the tooltip library and, thus, should probably own the tooltip property. I get there are some situations that will overlap (e.g. using a UI library that has tooltips built in) but let's implement the second part of your fix and reassess if it continues to be a pain. How does that sound?

coladarci:
To my knowledge, checking the constructor in JS is a dogey area - I have a simpler fix in a local branch - https://github.com/sir-dunxalot/ember-tooltips/pull/48 - pretty "hacky" but our prod site went down so I needed something quick :) Thoughts?

coladarci:
Another way I've handled this in the past is to add a property to your tooltip object, like isEmberTooltip, so you can replace this conditional w/ a check for that property. Still hacky but a little more explicit.. Thoughts?

sir-dunxalot:
I think either of those should work. The underlying library has a `Tooltip` class so I think we'd be ok there. But I defer to you if you know of any other issues checking the constructor. I'll test it out and if that doesn't work we'll use the second approach - I've used it in multiple projects before and it's been reliable.

coladarci:
Hey @sir-dunxalot - have you had a chance to check this out? Thanks!

sir-dunxalot:
Sorry, not yet. I'll try to do so before the weekend. Thanks for the reminder.

sir-dunxalot:
This should be fixed by #48, which is now on master and released. If you're still having issues we can reopen this issue.

PR #48, opened 2016-02-29
#47 - be careful to not assume that the tooltip defined is, in fact, …

coladarci:
…the type of tooltip we are expecting

### DIFF
--- a/addon/mixins/components/tooltips.js
+++ b/addon/mixins/components/tooltips.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 import renderTooltip from 'ember-tooltips/utils/render-tooltip';
 
 const { on } = Ember;
+const { Tooltip } = window;
 
 export default Ember.Mixin.create({
 
@@ -62,7 +63,7 @@ export default Ember.Mixin.create({
   destroyTooltip: on('willDestroyElement', function() {
     const tooltip = this.get('tooltip');
 
-    if (tooltip && Ember.typeOf(tooltip) === 'object') {
+    if (tooltip && tooltip instanceof Tooltip) {
       tooltip.effect(null); // Remove animation
       tooltip.detach();     // Remove the tooltip from the document
       this.$().off();       // Remove all event listeners


sir-dunxalot:
I'm just seeing this sorry.

Would a slightly more robust approach be to check if the constructor of `tooltip` is this library's `Tooltip` class?

coladarci:
I tend to avoid that in JS, though if my concern is outdated, let me know. In any other language, hats exactly what you'd do... See this question for more context http://stackoverflow.com/questions/7039260/can-i-use-constructor-name-to-detect-types-in-javascript

sir-dunxalot:
I believe `instanceof` is pretty safe these days. I'd probably advocate for that approach instead of checking for the `detach` method.

addon/mixins/components/tooltips.js
```diff
@@ -62,7 +63,7 @@ export default Ember.Mixin.create({
   destroyTooltip: on('willDestroyElement', function() {
     const tooltip = this.get('tooltip');
 
-    if (tooltip && Ember.typeOf(tooltip) === 'object') {
```
coladarci: @sir-dunxalot  - take a look - I've updated this pull to do the `instanceof` check you suggested. It's worth noting that it looks like another attempt at fixing this bug was merged here https://github.com/sir-dunxalot/ember-tooltips/commit/432a943795c765f39a559dbe4421771eb41f9a99#diff-6d875f075dad177398e49011cc51f439R65 but that is simply not a good enough check; the collision in my case is w/ another object. (For others who might be reading this, https://github.com/Glavin001/ember-c3 uses a property `tooltips` - (a hash, therefore also an `object`) for it's graphs, so just by adding this tooltip library, any time you navigate away from a graph, your app crashes since this library is added to every component and hooks into the `willDestroyElement`). 

I've tested this w/ my app and it works for me, would love very much to get a release of this! 

Also, I tried to write a failing test for for whatever reason can't run your tests;
```
The Broccoli Plugin: [SassCompiler] failed with:
Error: "pre" failed to @extend "%relative".
The selector "%relative" was not found.
Use "@extend %relative !optional" if the extend should be able to fail.
    at options.error (/Users/gcoladarci/lib/ember-tooltips/node_modules/ember-cli-sass/node_modules/broccoli-sass-source-maps/node_modules/node-sass/lib/index.js:271:32)
```

coladarci:
Not sure what to do about the failing tests; the fail for me locally as well but can't imagine my change is responsible for a sass fail! Any thoughts?

addon/mixins/components/tooltips.js
```diff
@@ -62,7 +63,7 @@ export default Ember.Mixin.create({
...
+    if (tooltip && tooltip instanceof Tooltip) {
```
coladarci: @sir-dunxalot  - take a look - I've updated this pull to do the `instanceof` check you suggested. It's worth noting that it looks like another attempt at fixing this bug was merged here https://github.com/sir-dunxalot/ember-tooltips/commit/432a943795c765f39a559dbe4421771eb41f9a99#diff-6d875f075dad177398e49011cc51f439R65 but that is simply not a good enough check; the collision in my case is w/ another object. (For others who might be reading this, https://github.com/Glavin001/ember-c3 uses a property `tooltips` - (a hash, therefore also an `object`) for it's graphs, so just by adding this tooltip library, any time you navigate away from a graph, your app crashes since this library is added to every component and hooks into the `willDestroyElement`). 

I've tested this w/ my app and it works for me, would love very much to get a release of this!

sir-dunxalot: Great, thanks for the update. I'll get on this ASAP so we can make another release - likely sometime this week.

PR #49, opened 2016-03-01
Fix applying of typeClass as defined by vendor css

ohcibi:
The vendor's css defines type classes like `.tooltip.light` and not `.tooltip-light`.

This also fixes the way the class names are asserted. Previously a simple string match was performed which won't fail for a given classname of `foo-bar` and an expected class name of `foo`.

### DIFF
--- a/addon/utils/render-tooltip.js
+++ b/addon/utils/render-tooltip.js
@@ -100,7 +100,7 @@ export default function renderTooltip(domElement, options, context) {
     /* Prefix type class */
 
     if (typeClass) {
-      newOptions.typeClass = 'tooltip-' + typeClass;
+      newOptions.typeClass = typeClass;
     }
 
     /* Set the correct hide and show events */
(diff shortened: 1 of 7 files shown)


sir-dunxalot:
Sorry. I missed this PR. The `.tooltip-light` was originally added because the `.light` class was considered too generic and conflicted with the apps' stylesheets. It would appear that the not-updated stylesheet was an unintentional regression and I agree that we need a fix for it.

I'm going to see if we can port over the tooltip styling into this library so we can use some of this PR (like the better class name assertions) but I expect reverting to `.light` will cause stylesheet conflicts.

ohcibi:
I'm equally happy with both `.light` or `.tooltip-light`. I just wanted it to work and I simply decided **not** to fiddle with the vendors style sheets when I made this. But this fix is also possible by leaving the implementation and changing the style sheet. If you could upgrade the vendor styles I'd change this PR to only improve the class name check.

sir-dunxalot:
The issues here were stemming from the vendor library setup. This is no longer an issue with release `1.0.0` ([here](https://github.com/sir-dunxalot/ember-tooltips/releases/tag/1.0.0)) as that vendor library has been removed.

PR #52, opened 2016-03-25
Usage of `Ember.merge` is deprecated, use `Ember.assign` instead

Dhaulagiri:
Getting out in front of this deprecation that will start throwing warnings in Ember 2.5.

_Huh, well that wasn't as easy as I thought.  I'll circle back once I get a little more time_


sir-dunxalot:
No worries. You could always try using `$.merge` to avoid the deprecation warning. I believe the functionality is similar for this purpose.

PR #55, opened 2016-04-13
Skips JavaScript import if on FastBoot

locks:


sir-dunxalot:
I don't have any experience with fast boot but it looks like this will only import the vendor file if the app is running in fast book, no? Isn't that the opposite of what we want?

locks:
You are most correct, I forgot the negation operator! I'll update on a short while.

locks:
There we go. Can you squash through the GitHub interface?

sir-dunxalot:
Great, thanks. And yep - I'll squash in Github.

PR #64, opened 2016-05-19
Fix aria-hidden bug

samdemaeyer:
Yesterday was **Global Accessibility Awareness Day**, as this was the ember-meetup theme, we looked at some much used ember-plugins and checked if the accessibility was correct. We found a small bug with the tooltip. It does not read out the related tooltip on with the screenreader. 
`aria-hidden` value was inverted (**`true`** on show and **`false`** on hidden) and didn't allow user using screen reader to actually read the tooltip content.


sir-dunxalot:
Awesome, thanks for the fix!

PR #66, opened 2016-05-22
https://github.com/sir-dunxalot/ember-tooltips/issues/5 - fix - obser…

Kinjeiro:
…ver changing in tooltip-on-parent content

For observ changing we should delete component. We should move it to tooltip.
Tooltip can take the element of this component as content

### DIFF
--- a/addon/components/tooltip-on-parent.js
+++ b/addon/components/tooltip-on-parent.js
@@ -20,13 +20,16 @@ export default Ember.Component.extend({
       if (!parentView) {
         console.warn('No parentView found');
       } else if (parentView.renderTooltip) {
+        //Need to show for calculation correct position in Tooltip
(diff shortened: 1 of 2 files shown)


sir-dunxalot:
As noted in the other thread for #5, this is fixed on master branch by the `1.0.0` release.

We can still fix this on the `pre-1.0` branch ([here](https://github.com/sir-dunxalot/ember-tooltips/tree/pre-1.0)) and issue a patch as `0.7.1` if you think that is a good idea.

sir-dunxalot:
Sorry for the bad timing - it seems like we both tackled this at the same time!

sir-dunxalot:
I'm closing this for now. let me know if you think it should be reopened.

PR #70, opened 2016-05-29
render tooltip immediately in testing mode

cooperjbrandon:
When writing acceptance tests that involve ember-tooltips, the `didUpdate()` method in the `tooltip-on-element` method renders the tooltip 1000 ms later. However, if there are many tooltips on the page, this causes the acceptance test to take a very long time.

So in testing mode, this sets the default update time to 0ms.

### DIFF
--- a/addon/components/tooltip-on-element.js
+++ b/addon/components/tooltip-on-element.js
@@ -70,6 +70,7 @@ export default EmberTetherComponent.extend({
   classPrefix: 'tooltip',
   classNames: ['tooltip'],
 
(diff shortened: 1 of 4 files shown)


cooperjbrandon:
@sir-dunxalot yay or nay?

sir-dunxalot:
Sorry, Brandon, I completely missed this. I think this is a great idea. It will help a lot of people who create their own forks of this library. I will pull down and release this as soon as I can get to it!

cooperjbrandon:
is this good?

sir-dunxalot:
Yep, sorry. This is now on master and will be released with the next release.

PR #78, opened 2016-06-22
Fix aria-describedby syntax.

blimmer:
Previously, we were creating elements like this:
```html
<div aria-describedby='#ember123'>
```
but the aria spec expects the attribute value to be an id without the `#` like this:
```html
<div aria-describedby='ember123'>
```

I discovered this when looking at the Chrome accessibility tools:

#### Before
screen shot 2016-06-22 at 4 58 23 pm

#### After
screen shot 2016-06-22 at 5 00 48 pm

### DIFF
--- a/addon/components/tooltip-on-element.js
+++ b/addon/components/tooltip-on-element.js
@@ -292,7 +292,7 @@ export default EmberTetherComponent.extend({
     }
 
     $target.attr({
-      'aria-describedby': `#${this.get('elementId')}`,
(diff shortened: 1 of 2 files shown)


sir-dunxalot:
Sorry, I must have overlooked this. Good find!

I am rerunning the test suite but if it fails do you mind merging master branch into this again? That should fix the `ember-beta` failure issues. Once the suite is passing this PR will be good to merge!

sir-dunxalot:
Released as `2.0.0`.

Issue #79, opened 2016-06-29
rename tooltip class to ember-tooltip


andrew-zenefits:
We have a massive codebase and are migrating to ember-tooltips for all our tooltip needs. In the past people have created `tooltip` classes for their one-off tooltip components. This means that a plain inclusion of this library will introduce many style conflicts to existing tooltips, and vice-versa. Right now our options are...

1) Fork our own PR with a temporary fix (extended the components, additional className for specificity, and add this in the library's styles `.ember-tooltip.tooltip`). We would need to keep this fork until we've deprecated all of our existing `.tooltip` class reliant components and instance.
2) Upgrade this library to use `.ember-tooltip` classes everywhere instead of `.tooltip`. 

It seems that more specificity in naming conventions is better and that if this library were to use `.ember-tooltip` there would be less chance of similar roadblocks for large companies hoping to use it.

If this is an acceptable solution I can address it.

@MiguelMadero ^

sir-dunxalot:
I think 2) seems like an acceptable solution. This issue came up once before with #15. An `.ember-tooltip` class namespace would be a stronger solution to that issue than just `.tooltip` so I'm in favor!

This would be a backwards incompatible styling change but I think that's ok if we offer good release notes for the change.

MiguelMadero:
Thanks for tagging me. In addition to 2, I think this is a good emerging convention to have a one to one mapping between your component names and classes. If we decided to follow that `tooltip-on-element` may be a better name. 

Regardless of the name chosen, if we wanted to minimize the impact we could have two separate releases:

1. Do a patch release (1.0.1 if done now) with both classes, all of the styles would be on `tooltip-on-element`, but the `.tooltip` class will still be applied by default in case someone is styling that explicitly to target it instead. Unfortunately, it's hard to raise deprecation warnings for something like this and we can't assume everyone is using sasslint. So we would have to rely on release notes. 
2. Do a major release (2.0.0 if done now) with the same, but the flag turned-off or removed, ppl may still opt-in on the old behavior by explicitly adding the class if they don't want to refactor their css rules.

andrew-zenefits:
Regarding the 1-1 mapping of components and classnames, *all* of the styles are shared between the `tooltip-on-element` and the `tooltip-on-component`. If this is the case we should decide on a shareable class name. I feel like `ember-tooltip` is a pretty good solution but would be open to different names.

[...]

PR #80, opened 2016-06-29
replace "tooltip" classes with "ember-tooltip" classes

andrew-zenefits:
This addresses the issue that was raised in https://github.com/sir-dunxalot/ember-tooltips/issues/79

### DIFF
--- a/addon/components/tooltip-on-element.js
+++ b/addon/components/tooltip-on-element.js
@@ -121,7 +121,7 @@ export default EmberTetherComponent.extend({
   }),
 
   effectClass: computed(function() {
-    return `tooltip-${this.get('effect')}`;
+    return `ember-tooltip-${this.get('effect')}`;
   }),
 
   positionClass: computed(function() {
@@ -165,7 +165,7 @@ export default EmberTetherComponent.extend({
   typeClass: computed(function() {
     const type = this.get('type');
 
-    return type ? `embertooltip-${type}` : null;
+    return type ? `ember-tooltip-${type}` : null;
   }),
 
   /* Private CPs */
--- a/addon/components/tooltip-on-element.js
+++ b/addon/components/tooltip-on-element.js
@@ -67,8 +67,8 @@ export default EmberTetherComponent.extend({
   'aria-hidden': computed.not('tooltipIsVisible'),
   attributeBindings: ['aria-hidden', 'role', 'tabindex'],
   classNameBindings: ['effectClass'],
-  classPrefix: 'tooltip',
-  classNames: ['tooltip'],
+  classPrefix: 'ember-tooltip',
+  classNames: ['ember-tooltip'],
 
   _hideTimer: null,
   _showTimer: null,
@@ -124,7 +124,7 @@ export default EmberTetherComponent.extend({
   positionClass: computed(function() {
     const targetAttachment = this.get('targetAttachment');
 
-    return `tooltip-${targetAttachment.replace(' ', ' tooltip-')}`;
+    return `ember-tooltip-${targetAttachment.replace(' ', ' ember-tooltip-')}`;
   }),
 
   sideIsVertical: computed(function() {
@@ -162,7 +162,7 @@ export default EmberTetherComponent.extend({
   typeClass: computed(function() {
     const type = this.get('type');
 
-    return type ? `tooltip-${type}` : null;
+    return type ? `embertooltip-${type}` : null;
   }),
 
   /* Private CPs */
@@ -303,7 +303,7 @@ export default EmberTetherComponent.extend({
     let renderedSide;
 
     ['top', 'right', 'bottom', 'left'].forEach(function(side) {
-      if ($_tether.hasClass(`tooltip-target-attached-${side}`)) {
+      if ($_tether.hasClass(`ember-tooltip-target-attached-${side}`)) {
         renderedSide = side;
       }
     });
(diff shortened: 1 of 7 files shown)


andrew-zenefits:
@sir-dunxalot it is passing the `default`, `ember-1.13`, and `ember-release` test scenarios.

Can you review?

addon/components/tooltip-on-element.js
```diff
@@ -67,8 +67,11 @@ export default EmberTetherComponent.extend({
   'aria-hidden': computed.not('tooltipIsVisible'),
   attributeBindings: ['aria-hidden', 'role', 'tabindex'],
   classNameBindings: ['effectClass'],
-  classPrefix: 'tooltip',
-  classNames: ['tooltip'],
+  classPrefix: 'ember-tooltip',
+  classNames: ['ember-tooltip', 'tooltip'],
```
MiguelMadero: I think we should decide if the two classes solution is overkill or not and how this will be released. I'll defer to @sir-dunxalot that was only my suggestion.... 
That said, if we decide to go with 2 classes, I think we need to also apply two classes for postitionClass, effectClass and typeClass for the same reasons.... So given the complexity it may introduce, I'm leaning more towards not doing it. I think this deserves a major change, not strictly an "API change", but certainly a breaking change.

sir-dunxalot: I'm not opposed to the breaking change by any means so I would lean towards the simplicity of a single class (`.ember-tooltip`).

sir-dunxalot:
Looks good overall! Any idea what's causing the large number of failures on ember-beta? Maybe failures on ember's end?

andrew-zenefits:
Not sure what's causing the beta failures. I looked back at the prior few PRs and they all seemed to fail on beta so I'm pretty confident I dint introduce any new errors.

addon/components/tooltip-on-element.js
andrew-zenefits: I think .ember-tooltip would be fine

@miguelmadero why would we need "two classes for postitionClass, effectClass and typeClass for the same reasons...."?

sir-dunxalot:
Ok. I'll merge this into master now and will either remove beta from the prod build for now or will just release with failures.

sir-dunxalot:
I take that back - I see the inline comment hasn't been resolves yet. Will wait for that before merging.

addon/components/tooltip-on-element.js
MiguelMadero: > @miguelmadero why would we need "two classes for postitionClass, effectClass and typeClass for the same reasons...."?

That's not relevant anymore since it's simpler to have a single class and do it as a single release. *If* we were to do it in two releases and have 2 classes for the first release so it doesn't break in case someone is styling `tooltip`, then following the same principle, we would have to maintain two classes for `postionClass` et.al because someone might be installing `tooltip-target-attached-left` which is changing to `ember-tooltip-target-attached-left`. 

Anyway, none of this matter since it's simpler to have a single release.

andrew-zenefits: Thanks Miguel.

@sir-dunxalot I think we do run a risk of existing projects having already applied styles to `tooltip-target-attached-left`, although the risk of that is much smaller than the risk of them having applied styles to `.tooltip` (which we do have temporary backwards compatibility for). The question becomes do we want to support backwards compatibility for the edge-cases of `tooltip-*` classes.

I think we should add release notes about continuing, temporary support for the `tooltip` class but say that the `tooltip-*` classes will be deprecated in the release and should be addressed before upgrading.

andrew-zenefits:
@sir-dunxalot I think we have addressed the inline comment after we get your thoughts on it.

addon/components/tooltip-on-element.js
sir-dunxalot: I'm honestly fine switching straight from `.tooltip` to `.ember-tooltip` because there's no easy way enforce and warn about deprecations for classes. I don't have a problem issuing a breaking release for this (`2.0.0`).

Thus, I'd switch to classes like `.ember-tooltip-target-attached-left` and not support `.tooltip-target-attached-left`. I'd also remove `.tooltip` completely. However, if you want to support both I'm fine with that.

andrew-zenefits: I agree with this whole approach. I think that keeping `.tooltip` in adds some degree of backwards compatibility so it seems like an easy win.

andrew-zenefits: Come to think of it I don't think we should keep the `.tooltip` class. The CSS can mix in two ways.
* Using `.ember-tooltip-*` makes sure that this library's CSS doesn't affect the project (ie. Zenefits app) code.
* But if we keep `.tooltip` here then a projects' tooltip classes may unintentionally affect the style of this library's tooltips.

I agree that we should make a clean break and only use `.ember-tooltip-*` classes. We should add a release note saying that if people want to add additional style then they need to apply it to the `.ember-tooltips` class instead of `.tooltip`.

I'll make a new commit that addresses this, run the tests again, then I believe we're good to go.

andrew-zenefits:
Tests are passing just as before (failing on 2.7 and 2.7 canary)

sir-dunxalot:
Great, thanks for the good discussion and progress here. I will get this merged into master and will see what I can do to get a passing suite before releasing within the next fews days.

If this is a blocker, let me know and I can try and release before the weekend.

andrew-zenefits:
Thank you very much Duncan. This is not a blocker for us and we can wait till then.

Now on to the popover work

sir-dunxalot:
I will be releasing a `1.x` version now and once that's done this will be released as a major release.

sir-dunxalot:
Released as `2.0.0`. Thanks for all the great work that's gone into this!

PR #81, opened 2016-06-30
Add isDestroyed & isDestroying checks

ilucin:
I have a scenario where component was in isDestroying state while ember-tooltips was trying to:
- render new tooltip content
- remove the tooltip-on-parent component

It happens when tooltips are inside of items of ember-power-select component.

This PR fixes the issue. I'm not exactly sure why is this happening but I guess these checks can't hurt anyone. Please correct me if I'm wrong :)

I don't know if this is happening in v1.0, can't upgrade yet (because of breaking changes).

### DIFF
--- a/addon/components/tooltip-on-parent.js
+++ b/addon/components/tooltip-on-parent.js
@@ -25,7 +25,7 @@ export default Ember.Component.extend({
         console.warn('No renderTooltip method found on the parent view of the {{tooltip-on-parent}} component');
       }
 
-      if (!this.get('isDestroyed') && !this.get('isDestroying')) {
+      if (!this.isDestroying) {
         this.remove();
       }
     });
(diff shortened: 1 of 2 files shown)


addon/components/tooltip-on-parent.js
```diff
@@ -25,7 +25,9 @@ export default Ember.Component.extend({
         console.warn('No renderTooltip method found on the parent view of the {{tooltip-on-parent}} component');
       }
 
-      this.remove();
+      if (!this.get('isDestroyed') && !this.get('isDestroying')) {
```
sir-dunxalot: Do we need `isDestroyed`? Isn't `isDestroying` true when `isDestroyed` is true? Looking at the [Ember docs](http://emberjs.com/api/classes/Ember.CoreObject.html#property_isDestroying) I'm not sure. Let me know how you want to proceed and we can get this merged.

sir-dunxalot:
Note to myself: this is merging into `pre-1.0`.

addon/components/tooltip-on-parent.js
ilucin: Yeah, I guess you're right. There can't be a situation where isDestroying=false and isDestroyed=true. I'll update the PR

sir-dunxalot: Sounds good. Once that's done I'll merge and release as a patch!

PR #84, opened 2016-07-14
Moves ember-beta to the list of `allow_failures`

MiguelMadero:
Many builds are failing for this. While it is important to be aware of this failures and try to fix them in advance, I don't think they should necessarily be a blocker for CI and it might be better to have a separate issue to address these failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
 matrix:
   fast_finish: true
   allow_failures:
+    - env: EMBER_TRY_SCENARIO=ember-beta
     - env: EMBER_TRY_SCENARIO=ember-canary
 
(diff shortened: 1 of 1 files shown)


sir-dunxalot:
Sounds good to me - thanks for taking the time to think about this. Merging now.

I will come back to this to try and ascertain why ember-beta is failing so badly in the future.

MiguelMadero:
@sir-dunxalot thanks for merging it, it will help other PRs get a green build. I created https://github.com/sir-dunxalot/ember-tooltips/issues/87 as a reminder to fix this and make it easier to track it, I think @andrew-zenefits was also thinking of doing something about it.

PR #94, opened 2016-09-16
Fix: check for 'isDestoyed' when _showTimer runs

gkappen:
This fixes an issue where an exception would be thrown about calling set on a destoyed component. This can happen if a component that has a tethered tooltip is destroyed before the tooltip has been shown


sir-dunxalot:
Great, thank you!

PR #96, opened 2016-09-21
Disable tethering while tooltips are hidden

joukevandermaas:
This was causing performance issues on pages with many tooltips, because each repositioning triggered a forced reflow (which is expensive in most browsers, but especially in IE).

### DIFF
--- a/addon/components/tooltip-and-popover.js
+++ b/addon/components/tooltip-and-popover.js
@@ -360,8 +360,6 @@ export default EmberTetherComponent.extend({
   }),
 
   show() {
-    this.startTether();
-
     // this.positionTether() fixes the issues raised in
     // https://github.com/sir-dunxalot/ember-tooltips/issues/75
     this.positionTether();
@@ -393,6 +391,7 @@ export default EmberTetherComponent.extend({
 
       const _showTimer = run.later(this, () => {
         if (!this.get('destroying') && !this.get('isDestroyed')) {
+          this.startTether();
           this.set('isShown', true);
         }
       }, delay);
@@ -402,6 +401,7 @@ export default EmberTetherComponent.extend({
 
       /* If there is no delay, show the tooltop immediately */
 
+      this.startTether();
       this.set('isShown', true);
     }
 
--- a/addon/components/tooltip-and-popover.js
+++ b/addon/components/tooltip-and-popover.js
@@ -85,15 +85,20 @@ export default EmberTetherComponent.extend({
 
   /* Properties */
 
-  attributeBindings: ['aria-hidden', 'role', 'tabindex'],
+  attributeBindings: ['aria-hidden', 'role', 'tabindex', 'data-tether-enabled'],
   classNameBindings: ['effectClass'],
   classPrefix: 'ember-tooltip-or-popover',
-
+  
   _didUpdateTimeoutLength: 1000, // 1000 ms or 0 ms, depending whether in test mode
   _hideTimer: null,
   _showTimer: null,
+  _isTetherEnabled: true,
 
   /* CPs */
+
+  'data-tether-enabled': computed('_isTetherEnabled', function() {
+    return this.get('_isTetherEnabled') ? 'true' : 'false';
+  }),
 
   'aria-hidden': computed('isShown', function() {
     return this.get('isShown') ? 'false' : 'true';
@@ -240,6 +245,8 @@ export default EmberTetherComponent.extend({
 
     this.set('isShown', false);
     this.sendAction('onHide', this);
+
+    this.stopTether();
   },
 
   didInsertElement() {
@@ -298,6 +305,10 @@ export default EmberTetherComponent.extend({
     }
 
     this.set('offset', offset);
+
+    if (!this.get('isShown')) {
+      this.stopTether();
+    }
   },
 
   /*
@@ -327,6 +338,7 @@ export default EmberTetherComponent.extend({
     const isShown = this.get('isShown');
 
     if (isShown) {
+      this.startTether();
       const duration = cleanNumber(this.get('duration'));
 
       run.cancel(this.get('_hideTimer'));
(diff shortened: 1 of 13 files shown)


joukevandermaas:
I can't figure out why the build is failing, but I'm pretty sure it's not (entirely) related to this PR. Can someone maybe point me in the right direction?

addon/components/tooltip-on-element.js
```diff
@@ -336,6 +338,10 @@ export default EmberTetherComponent.extend({
     }
 
     this.set('offset', offset);
+
+    if (event !== 'none' || this.get('tooltipIsVisible') !== true) {
```
a15n: nit: `!this.get('tooltipIsVisible')`

addon/components/tooltip-on-element.js
```diff
@@ -430,6 +440,16 @@ export default EmberTetherComponent.extend({
     this.sendAction('onTooltipShow', this);
   },
 
+  stopTether() {
+    run.schedule('afterRender', this, function() {
```
a15n: nit: if you use arrow functions you won't need to pass in `this`

```
run.schedule('afterRender', () => {
  ...
});
```

addon/components/tooltip-on-element.js
```diff
@@ -430,6 +440,16 @@ export default EmberTetherComponent.extend({
...
+      this.get('_tether').disable();
+    });
+  },
+
+  startTether() {
+    this.get('_tether').enable();
+  },
```
a15n: could you add tests to check that the tether is enabled when we expect it to be?

```
isTetherEnabled: computed.alias('_tether.enabled'),
```

andrew-zenefits: We copied your fix here at Zenefits and did this to get test coverage

https://github.com/zenefits/ember-tooltips/pull/21/files#diff-eafc17148ee89489b3d8098122b506e6R409

sir-dunxalot:
As Andrew said, this is a great fix. Good catch and thank you for putting in the time to issue the PR. I will wait until #94 has landed to merge this but the code looks good so I don't have any comment there. This will likely need a quick rebase to resolve the merge conflicts once #94 is merged onto master.

joukevandermaas:
OK, I'll address @a15n's comments and rebase after #94 is merged, then. Thanks!

a15n:
@joukevandermaas I think this may be helpful. It has full test coverage. I don't know if what I did with an `is-tether-enabled` attribute is ideal. A 'js-' prefixed class may be better. 

https://github.com/zenefits/ember-tooltips/compare/v2.0.0-zenefits.1...v2.0.0-zenefits.2

joukevandermaas:
I see some refactoring happened in the meantime, while adding popovers. I hope to get this PR fixed up today or tomorrow if I can.

sir-dunxalot:
Sounds great, Jouke! Let myself or @a15n know if you have any questions about the refactoring from popovers.

joukevandermaas:
I rebased, adapted the code and added some asserts to the visibility tests. I squashed too, so the 'old' code is no longer visible here.

I don't know why the beta build fails, but it doesn't look like it's anything to do with this PR.

tests/integration/components/tooltip-is-visible-test.js
```diff
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 import { moduleForComponent, test } from 'ember-qunit';
-import { assertHide, assertShow } from '../../helpers/sync/assert-visibility';
+import { assertHide, assertShow, assertTetherEnabled, assertTetherDisabled } from '../../helpers/sync/assert-visibility';
```
a15n: I would recommend against creating a new assertion just for the tether. This means that people will always have to opt in to the usage and will likely skip that test. Instead I would add the check as part of assertShow and assertHide. When I did this I found some holes in my enable/disable coverage

https://github.com/zenefits/ember-tooltips/pull/21/files#diff-c2e9fd979e3b24030663b2c4c84cd5beR3

addon/components/tooltip-and-popover.js
```diff
@@ -419,4 +435,20 @@ export default EmberTetherComponent.extend({
     this.sendAction('onDestroy', this);
   },
 
+  startTether() {
+    // can't depend on `_tether.enabled` because it's not an
+    // Ember property (so won't trigger cp update when changed)
```
a15n: good note

tests/integration/components/tooltip-is-visible-test.js
a15n: I think this should be blocking

a15n:
the two test failures are expected and not related to these changes. @sir-dunxalot maybe we should make the `ember-beta` test failure allowable?

tests/integration/components/tooltip-is-visible-test.js
joukevandermaas: Yeah, good point. I'll try split it (or make a new PR if you want to merge already).

lvl99:
Pretty keen to see this integrated with the main build. Currently have some extreme lag after adding tooltips to my Ember app.

sir-dunxalot:
Yes, I think making `ember-beta` failure allowable is acceptable given the conversations about the impact of glimmer (see #97). As such, this PR should be good to go once @a15n 's code comments have been addressed.

joukevandermaas:
@a15n I moved the tethering assert into the existing helper. There was a slight complication, because for delayed tooltips the tethering should be enabled but the tooltip should remain hidden for a while. I added some flag to the helper now, but another potential solution is to split this specific scenario into a new helper. Let me know what you think.

a15n:
@joukevandermaas thanks for your work on this.

When we did this for our local repo we didn't have to provide any additional parameters for the tests. I think that if you add the `startTether()` call inside the if/else within the show function you won't need the additional test parameter or special handling. 

https://github.com/zenefits/ember-tooltips/pull/21/files#diff-eafc17148ee89489b3d8098122b506e6R382

joukevandermaas:
Good point, I've updated the code to do that and the special assert is no longer needed. Thanks!

a15n:
Great work @joukevandermaas! This code LGTM.

@sir-dunxalot ^

sir-dunxalot:
+1. I will merge this, move `ember-beta` to allowed failures, and will release today with the passing test suite.

Great work all.

sir-dunxalot:
Releases as `2.3.1`.

PR #97, opened 2016-10-03
Beta & canary fixes

a15n:
Master currently fails in ember-beta and ember-canary ~40 times each. The cause of these failures are....

* `this.get('parentView')` returns `null`. [This ticket](https://github.com/emberjs/ember.js/issues/11170) mentions similar issues. I believe that they continued support for parentView but it's possible we're not referencing the most up to date ember-canary?
  * fix: use `tooltip-on-element` everywhere but the tooltip-on-component test (this test is still failing)
* in canary and beta `onTooltipRender` is called twice after the tooltip is destroyed ([notes](https://github.com/sir-dunxalot/ember-tooltips/pull/97/files#diff-7bc9c93b5304eef5520c3862a4080311R80)) 
  * fix: remove the assert.expect(10)
* some tests were failing due to a short timeout
  * fix: increase the timeout

Result: 1 test failure each instead of 40.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,15 +1,12 @@
 {
   "name": "ember-tooltips",
   "dependencies": {
-    "ember": "components/ember#canary",
+    "ember": "^2.6.0",
     "ember-cli-shims": "0.1.1",
     "ember-cli-test-loader": "0.2.2",
     "ember-qunit-notifications": "0.1.0",
     "hammer.js": "^2.0.8",
     "matchMedia": "0.2.0",
     "es5-shim": "^4.5.8"
-  },
-  "resolutions": {
-    "ember": "canary"
   }
 }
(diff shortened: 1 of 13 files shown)


a15n:
@sir-dunxalot I've addressed some of the failing tests here. https://github.com/sir-dunxalot/ember-tooltips/pull/97

Let me know if I can help in any other way to land this.

sir-dunxalot:
I haven't been able to get beta and canary passing. I believe Glimmer, which is introduced in ember 2.9, is causing these failures. I haven't found any documentation or issues to point me in the direction of fixing the remaining tests. At this point, I suspect the failures may be because ember testing libraries are not compatible with Glimmer. I'll release with the failing test and will circle back when 2.9 is releases with update instructions for developers.

a15n:
Good plan. Thanks

Issue #82, opened 2016-07-01
Positional options lost in 1.0


bilherron:
As I was up upgrading this (fantastic!) addon, it looks like we have lost some granularity in positioning the tooltip (ie top-left, top-right, bottom-left, bottom-right). I think ember-tether can support these, and I'm happy to take a crack at implementing it in ember-tooltips but I wanted to check in first on your thoughts @sir-dunxalot. Thanks!

sir-dunxalot:
Agreed. I'd love to put them back in. They were removed for simplicity to get 1.0 out the door and I fully support reimplementing them.

We will need to make sure that both side and [attachment](https://github.com/sir-dunxalot/ember-tooltips/blob/master/addon/components/tooltip-on-element.js#L78) function correctly with regards to what ember-tether expects.

Also, [position class](https://github.com/sir-dunxalot/ember-tooltips/blob/master/addon/components/tooltip-on-element.js#L124) should support these new multi-word positions.

Also, see [this bit of code](https://github.com/sir-dunxalot/ember-tooltips/blob/master/addon/components/tooltip-on-element.js#L305-L335), which is a little hacky but presents options in a format that ember-tether expects.

Tests can be added to [the side integration test](https://github.com/sir-dunxalot/ember-tooltips/blob/master/tests/integration/components/side-test.js).

sir-dunxalot:
I also wonder if `position` is a better property name than `side`?

sir-dunxalot:
Hey @bilherron, I don't suppose you've been able to take a crack at this?

bilherron:
Haven't had a chance yet, sorry! Still on my radar, though.

renatocarvalho:
To me, this change was a big regression. Darsain tooltip is great because it calculates everything automatically. Now everything in breaking on my app. I like the idea of keeping things simple, but in order to make tooltips just work, it needs a lot of work. And that's what Darsain Tooltip did.

@sir-dunxalot Thanks for all your hard work though.

sir-dunxalot:
Thanks for the feedback @renatocarvalho! It's been a little rocky but it will work out in the end. :)

a15n:
@jcward10 has https://github.com/sir-dunxalot/ember-tooltips/pull/103 that would introduce more positional options. His PR introduces a `defaultPosition` which acts like a secondary position (if I understand correctly)...

```
{{#tooltip-on-component side='right' defaultPosition='top'}}
  template block text
[...]

PR #103, opened 2016-10-19
Allow for the tooltips defaultPosition to be changed

jcward10:
We have a use case where we want to be able to modify the verticalPosition of a side='right' tooltip.


a15n:
Thanks for making this @jcward10. There is an ongoing conversation about positional parameters in https://github.com/sir-dunxalot/ember-tooltips/issues/82. Could you review the conversation there?

jcward10:
Yeah definitely @a15n

sir-dunxalot:
@jcward10 Are you expecting to get back to this? No problem if not - I'd just like to know if this is something myself or @a15n should expect to pick up in the future. A PR from yourself would still be very valuable.

jcward10:
@sir-dunxalot I can definitely finish up this PR. I was wait for a response on #82. I'll review the comments on #82 and refactor this PR as appropriate. Hopefully, I'll have time on my lunch break today.

jcward10:
@a15n @sir-dunxalot What do we want to name the secondary positioning variable ? Right now it is defaultPosition based on an old const. Do we want to call it `secondaryPosition` or `relativeSidePosition` ?

a15n:
I'd prefer `relativeSidePosition`. 

I think the biggest challenge with this will be how the `relativeSidePosition` positions the tooltip and pointer in relation to tooltipTarget. I outlined some of the challenges [here](https://github.com/sir-dunxalot/ember-tooltips/issues/82#issuecomment-255497816) but there are other edge cases as well. Unfortunately we don't have reliable visual tests (tried to get ember-percy working but it didn't play well with our absolutely positioned tooltips) so it'll require a lot of "refresh-the-page" checks.

Thanks for looking into this!

jcward10:
@a15n & @sir-dunxalot I think I have it working now. It looks fine on the dummy/index.hbs, but I'm having a hard time to get 2 test to pass. They are in tests/integration/components/relative-side-position-test.js. In the tests, the tooltip center is suppose to be to the left of the targetCenter, but its not 😿. Do y'all have any idea why it would position correctly in the UI but not in the tests?

a15n:
Hey @jcward10 I checked out the branch and stress tested it locally. I was able to produce these outcomes which I think will need to be addressed first. I think the problem is the the tooltip is "_above and to the right_" (for example) instead of "_above and mostly to the right_" which would have been centered. There are many use cases where a tooltip is attached to a tiny icon so we'll need the pointer to be exactly positioned at the center, not slightly off center.

screen shot 2017-03-30 at 3 56 02 pm
screen shot 2017-03-30 at 3 54 49 pm

basicBrogrammer:
I'll take a look at it next week. I have my 1 year anniversary trip this week so I won't have to take a look.

sir-dunxalot:
Please note that the `master` branch will receive backwards-incompatible changes within a few days because ember-tooltips `3.0.0` is being released. This change will make it very hard to merge this PR.

To answer your question @basicBrogrammer, ideally the tooltip arrow does not sit outside the vertical and horizontal borders of the button. In other words, the example above would have the arrow stay entirely within the green area:

image

However, after re-reading this thread with fresh eyes, I am inclined to suggest that, in the interest of simplicity, this not be a property but be solved by CSS alone. Maybe this would be as simple as including instructions in the Readme or demo app for how to move the tooltip relatively.

sir-dunxalot:
Yes, a PR rewrite sounds good (as a feature or simple documentation change). `master` will be updated to `3.0.0` on the 5th October so that's when you can expect this PR to be unmergable.

@maxfierke - What do you think about making this an option vs documenting the CSS workaround?

maxfierke:
@sir-dunxalot @basicBrogrammer If I'm understanding this correctly, this might already be an option via popper.js using placement variants (which is where `side` goes in ember-tooltips 3.x) https://popper.js.org/popper-documentation.html#Popper.placements

If that's the case, I'd argue we should just ensure that ember-tooltips 3.x is setup to handle the variants (also mentioned in #253). Otherwise, IMHO this should be addressed with CSS.

sir-dunxalot:
@basicBrogrammer [Here is an example of the placement variants](https://popper.js.org/index.html#example4). Does this property support your use case AND are you (once stable) willing to update to ember-tooltips `3.0.0` (currently `3.0.0-beta.3`)? Three options exist for the path forward here:

- If the answer is yes to both, we can rely on placement variants and the PR can be closed
- If you need this functionality in ember-tooltips `2.x`, a documented workaround seems like the least time-intensive solution for this PR (given that `2.x` will soon be out-of-date)
- If you are able to update to ember-tooltips `3.0.0` in the near future and placement variants does not support your needs documented CSS workaround seems like the best solution for this PR

basicBrogrammer:
@sir-dunxalot I'm I think the popper.js placement is exactly what we were looking for! 👍 We should prolly document it. (if its not already) I'll upgrade to 3.0 and remove our hack to see what the diff is. I'll report back with what I find 👍 Thanks @maxfierke for the suggestion!

sir-dunxalot:
Great! Glad to hear it.

#282 (currently on `3.x`, soon to be on `master`) adds documentation for popper variants.

basicBrogrammer:
@sir-dunxalot I don't want to open a new issue but maybe I can to discuss it. Is there a good "click off" pattern that you know off?

sir-dunxalot:
@basicBrogrammer I have sent you a message on the Ember discord. We can go into details there!

Issue #100, opened 2016-10-14
Initial Rendering Performance Issues


a15n:
In our app we use tooltips a lot. On some pages there may be dozens of tooltips at a time (sometimes more if we use tables). We saw performance issues when a page or block of code was initially rendered with many tooltips within it because it would lag while each tooltip positioned and rendered itself, even if the tooltips were initially invisible.

In our app we also created a super simplified way to put in a tooltip called {{z-tooltip}}. This encapsulates an info icon and the corresponding tooltip component. You could pass any ember-tooltips related params to the wrapper component and it would pass them through to the `tooltip-on-component` component. 

<img width="696" alt="screen shot 2016-10-14 at 8 04 48 am" src="https://cloud.githubusercontent.com/assets/7050871/19392270/efa5ddb0-91e4-11e6-8a3a-145c43258eba.png">

We fixed the slow load times by "lazy loading" the `tooltip-on-element` within {{z-tooltip}}. Essentially we delay the insertion, rendering, and positioning of the tooltip until someone interacts with the target element (the question icon). We did this with some complex interaction handling and interaction mimicking after the tooltip was rendered. The result is a `tooltip-on-element` that only renders itself after the user has interacted with the icon using a mouseEnter, click, or focusin.

```
import Ember from 'ember';
import template from '../templates/components/z-tooltip';

export default Ember.Component.extend({
    tooltipText: null,
    tooltipClassNames: null,

    // default ember-tooltip options
    delay: 0,
    delayOnChange: true,
    duration: 0,
    effect: 'slide', // fade, slide, none
    event: 'hover', // hover, click, focus, none
    hideOn: null,
    role: 'tooltip',
    side: 'top',
    showOn: null,
    spacing: 10,
    tabindex: '0', // A positive integer (to enable) or -1 (to disable)
    isShown: false,
    keepInWindow: true,

    tagName: 'span',
    template: template,

    _shouldRenderTooltip: Ember.computed.or('isShown', 'hasUserInteracted'),

    willDestroyElement() {
        this._super(...arguments);

        const $tooltipIcon = this.$().find('.js-z-tooltip-icon');
        $tooltipIcon.off('mouseenter click focusin');
    },

    hasUserInteracted: false,
    actions: {
        recordUserInteraction(event) {
            if (!this.get('hasUserInteracted')) {
                this.set('hasUserInteracted', true);

                Ember.run.next(() => {
[...]

PR #104, opened 2016-10-22
A15n/lazy render

a15n:
### Overview
This introduces an opt-in for lazy rendering of tooltips and popovers. When `enableLazyRendering=true` a tooltip will only be rendered when the user interacts with the `$target` element *or* when `isShown=true`. This allows tooltip rendering to be delayed until when/if a user needs the tooltip, which results in significant load time improvements for pages with many tooltips and popovers. (see https://github.com/sir-dunxalot/ember-tooltips/issues/100)

### Notable additions
* [lazy-render-wrapper](https://github.com/sir-dunxalot/ember-tooltips/pull/104/files#diff-8096b1ffac251e79ea18b7393d9f241fR1) is a tagless component which will handles rendering and passes through `PASSABLE_OPTIONS`
* [tether-tooltip-and-popover](https://github.com/sir-dunxalot/ember-tooltips/pull/104/files#diff-aab8720cdae20e1f53c656a7aa537184R30) receives the `PASSABLE_OPTIONS` through on `didReceiveAttrs`.
* added new tests
* updated the dummy app and readme

### Inheritance Structure
`lazy-render-wrapper` --extends to--> `tooltip-on-element` --which renders--> `tether-tooltip-on-element`

Popovers follow the same inheritance structure. The `tether-` components lazily render after the user interacts.

### Release Schedule
I recommend we release this as version 2.4.0. This PR keeps `enableLazyRendering=false`. In 3.0.0 we should switch that to `true`. In the meantime people can experiment with lazy rendering on a per-component basis or with [component extension](https://github.com/sir-dunxalot/ember-tooltips#setting-defaults).

### Demo
Note that on page load there is only 1 tooltip/popover. As the user interacts the tooltips are rendered. By the end there are 9 tooltips/popovers in the page.

lazy tooltips


addon/components/tooltip-on-element.js
```diff
@@ -1,76 +1,54 @@
...
-
-        /* If the event is click, we don't want the
-        click to also trigger focusin */
-
-        if (event !== 'click') {
-          $target.focusin(() => {
-            this.show();
-          });
-        }
+		if (this.get('hasUserInteracted')) {
```
a15n: TODO change this to enableLazyRendering

addon/components/lazy-popover-on-element.js
```diff
@@ -0,0 +1,126 @@
...
+
+export default TooltipAndPopoverComponent.extend({
+
+  /* Options */
+  hideDelay: 250,
+
+  /* Properties */
+  classNames: ['ember-popover'],
+  _isMouseInside: false,
+  didRender() {
```
a15n: this is the only change in this file. The rest is a `git mv`.

addon/components/lazy-tooltip-on-element.js
```diff
@@ -0,0 +1,76 @@
+import Ember from 'ember';
```
a15n: result of a `git mv`

tests/integration/components/popover/hover-test.js
```diff
@@ -33,15 +33,15 @@ test('Popover: hover target, hover elsewhere', function(assert) {
...
-  }, 300);
+  }, 200);
 
   assert.expect(8);
 
 });
 
 test('Popover: hover target, hover popover (too slow)', function(assert) {
 
-  this.render(hbs`{{popover-on-element event="hover" hideDelay="250"}}`);
```
a15n: 250 is default

addon/components/lazy-popover-on-element.js
```diff
@@ -0,0 +1,129 @@
+import Ember from 'ember';
```
a15n: no changes to this file except the addition of the `didRender`. Appears changed due to `git mv`.

addon/components/lazy-tooltip-on-element.js
a15n: no changes to this file. Appears changed due to `git mv`.

tests/integration/components/popover/hover-test.js
```diff
@@ -33,15 +33,15 @@ test('Popover: hover target, hover elsewhere', function(assert) {
...
+  this.render(hbs`{{popover-on-element event="hover"}}`);
```
a15n: 250 is default

a15n:
@sir-dunxalot please review

@MiguelMadero, @SirZach, @lvl99 FYI

SirZach:
That's a lot of code I'm glad I didn't have to write :)

README.md
```diff
@@ -31,6 +32,10 @@ Version 1.0.0 removed <a href="http://darsa.in/tooltip/" target="_blank">darsain
 
 You can use and see the pre-1.0 version on [this branch](https://github.com/sir-dunxalot/ember-tooltips/tree/pre-1.0). Alternatively, install `"ember-tooltips": "0.7.0"` in your `package.json`.
 
+## 2.4.0 Release
+
+Version 2.4.0 introduces lazy rendering. Tooltips and popovers generally don't need to be rendered until the user has interacted with the `$target` element. Adding `enableLazyRendering=true` to your component will enable this future. In version 3.0.0 `enableLazyRendering` will default to `true` and you'll be able to opt-out of lazy rendering as necessary.
```
sir-dunxalot: "will enable this future" <-- Should read "will enable this feature"

addon/components/lazy-popover-on-element.js
sir-dunxalot: Understood. Thanks for the clarification.

addon/components/lazy-popover-on-element.js
```diff
@@ -0,0 +1,129 @@
...
+  classNames: ['ember-popover'],
+  _isMouseInside: false,
+
+  didRender() {
+    // the lazy-render popover component instance needs access to the childView
+    // so that it can call the childView's hide action
+    this._super(...arguments);
+
+    const parentView = this.get('parentView');
+    parentView.set('childView', this);
```
sir-dunxalot: I assume there's not a more Emberish way to do this? Maybe by passing an action into the component.

Also, is `childView` too generic here? What si the parent view already has a `childeView` property?

addon/components/lazy-render-wrapper.js
```diff
@@ -0,0 +1,149 @@
...
+const { computed, get, run, $ } = Ember;
+
+// https://github.com/emberjs/rfcs/issues/168
+export default function getParent(view) {
+  if (get(view, 'tagName') === '') {
+    // Beware: use of private API! :(
+    if (Ember.ViewUtils && Ember.ViewUtils.getViewBounds) {
+      return $(Ember.ViewUtils.getViewBounds(view).parentElement);
+    } else {
+      return $(view._renderNode.contextualElement);
```
sir-dunxalot: How does this code work with the changes incoming from Glimmer (Ember 2.9.0)?

sir-dunxalot: Actually, it looks like [the mentioned issue](https://github.com/emberjs/rfcs/issues/168) answers that question. Could you also link to https://github.com/emberjs/ember.js/pull/12500 in the code comment here?

addon/components/popover-on-element.js
```diff
@@ -1,122 +1,14 @@
-import Ember from 'ember';
-import TooltipAndPopoverComponent from 'ember-tooltips/components/tooltip-and-popover';
```
sir-dunxalot: Am I correct in believing a component no longer exists at this path? If developers have used this component for their own JS or styling then does this create a breaking change?

sir-dunxalot: I see below that this file exists. Maybe you could just explain how the architecture of the components has changed with this PR? For example, what inherits from what, what wraps what, why existing components can be used outside of the lazy rendering wrapper, etc.

addon/components/tooltip-and-popover.js
```diff
@@ -27,6 +27,16 @@ function cleanNumber(stringOrNumber) {
 
 export default EmberTetherComponent.extend({
 
+  passedPropertiesObject: null,
+  setPropertiesWithpassedPropertiesObject: on('didReceiveAttrs', function() {
```
sir-dunxalot: Minor point but `passed` is not camelCased in this method name.

addon/components/lazy-render-wrapper.js
```diff
@@ -0,0 +1,149 @@
...
+
+		if (this.get('_shouldRender')) {
+			// if the tooltip _shouldRender then we don't need
+			// any special $parent event handling
+			return;
+		}
+
+		const $parent = getParent(this);
+
+		INTERACTION_EVENT_TYPES.forEach((eventType) => {
```
sir-dunxalot: Can you please explain how this array is used elsewhere and why.

addon/templates/components/popover-on-element.hbs
```diff
@@ -1,3 +1,12 @@
-{{yield (hash
-	hide=(action "hide")
-)}}
+{{#if _shouldRender}}
```
sir-dunxalot: Correct me if I'm wrong:

- `_shouldRender` will be `false` by default if `enableLazyRendering` is `true`
- `_shouldRender` will be `true` by default if `enableLazyRendering` is `false`

tests/integration/components/actions-test.js
```diff
@@ -12,11 +12,12 @@ test('It supports deprecated lifecycle actions', function(assert) {
   // onTooltip____ actions are deprecated in favor of on on_____ actions
   // these actions will be supported until v3.0.0
 
+  // actions can have unique names
```
sir-dunxalot: What's the reason for changing the action names here? (just curious)

tests/integration/components/actions-test.js
```diff
@@ -135,24 +144,45 @@ test('It calls lifecycle actions', function(assert) {
...
 
 });
+
+test('It supports lifecycle closure actions with multiple arguments', function(assert) {
+
+  assert.expect(1);
+
+  let onRenderPassword;
+
+  this.on('onRenderFoo', (trickPassword, realPassword) => {
```
sir-dunxalot: Could you add a comment about what this test is doing.

tests/integration/components/effect-test.js
```diff
@@ -0,0 +1,20 @@
+import { moduleForComponent, test } from 'ember-qunit';
```
sir-dunxalot: Was this test simply missing before?

addon/templates/components/popover-on-element.hbs
```diff
@@ -1,3 +1,12 @@
...
+	{{#lazy-popover-on-element
```
sir-dunxalot: How does this `lazy-popover-on-element` affect the DOM?

README.md
```diff
@@ -359,6 +373,7 @@ import TooltipOnElementComponent from 'ember-tooltips/components/tooltip-on-elem
 export default TooltipOnElementComponent.extend({
   effect: 'fade',
   side: 'bottom',
+  enableLazyRendering: true,
```
sir-dunxalot: Should this change still be made in `your-app/components/tooltip-on-element}}` as the code example shows?

addon/components/popover-on-element.js
sir-dunxalot: I guess what I'm really trying to get to the bottom of is why we need the new `lazy-render-wrapper` component to achieve lazy loading of components.

sir-dunxalot:
1st review done. Overall this looks like a really sweet addition to the library. I'm very excited to get this in production!

README.md
a15n: No, I don't want to switch lazy rendering on by default until version 3.0.0.

I'm just showing it here to illustrate how to opt-in to this change for all of their tooltip/popover components.

addon/components/popover-on-element.js
a15n: I will update the description with a clearer example

addon/templates/components/popover-on-element.hbs
a15n: Correct. But let me clarify the first bullet point...

- `_shouldRender` will be `false` by default if `enableLazyRendering` is `true`
  * `_shouldRender` will be `true` by default if `enableLazyRendering=true` **and** `isShown=true`

addon/templates/components/popover-on-element.hbs
a15n: The `lazy-popover-on-element` is the Tether component. When this is rendered it will insert the Tether component, which will also attach the tether behavior to the target.

I'm unhappy with the name `lazy-popover-on-element`. I think I'm going to go with `popover-tether-component` and `tooltip-tether-component` to indicate the roles of the various components

tests/integration/components/actions-test.js
a15n: I added unique action names because when I initially made my changes compliant with this test I wrote a solution dependent on the action names being `onShow`, `onHide`, etc.

tests/integration/components/effect-test.js
a15n: correct

addon/components/lazy-render-wrapper.js
a15n: done

addon/components/lazy-render-wrapper.js
a15n: done

tests/integration/components/actions-test.js
a15n: done

sir-dunxalot:
@a15n This gets a +1 from me. Are you ok to make releases now? I'm happy to do it but if you want we just need to make sure the following happens:

- [ ] Bump version in `package.json` and commit
- [ ] Create Github tag
- [ ] Push tag to `master`
- [ ] Release on NPM

a15n:
@sir-dunxalot I am ready to release as 2.4.0 right now. I'll update the TODOs you listed as they're completed

MiguelMadero:
You could use ember release to do all of those at once.

(I didnt review the full changes, but Im a little bit worried about the
breaking changes).

sir-dunxalot:
Thanks @MiguelMadero. I use `release-it` locally but that is an interesting option.

I haven't noticed any breaking changes since `2.4.0` - have you seen anything in particular or are just weary because of the number of lines changed?

a15n:
Here is one breaking change, and the PR to address it https://github.com/sir-dunxalot/ember-tooltips/pull/110

addon/components/tether-tooltip-and-popover.js
```diff
@@ -27,6 +27,16 @@ function cleanNumber(stringOrNumber) {
 
 export default EmberTetherComponent.extend({
 
+  passedPropertiesObject: null,
+  setPropertiesWithPassedPropertiesObject: on('didReceiveAttrs', function() {
+    this._super(...arguments);
+
+    let passedPropertiesObject = this.get('passedPropertiesObject');
+    if (passedPropertiesObject) {
+      this.setProperties(passedPropertiesObject);
```
MiguelMadero: I'm not sure if this is what we want. You're just setting the properties, but not creating a binding. 

let's say we have prop1 as a passedProperty. Then if you change prop1 on the context, it will flow to the target, but it won't flow from the target to the `tether-tooltip-and-popover` since didReceiveAttrs won't be called again, so the actual component will be left with the old value. 

Can you test it? This may not be an issue, but if I'm right, it will be a bad way to break... 

In case it's an issue, it may be as simple as setting adding the same props with a `Binding` suffix or explicitly create bindings between wrapper and wrappee for each property on passedPropertiesObject. 

Also, I'm wondering if it's better to generalize this use case and apply it here, it seems like the "transparent wrapper" or "transparent proxy" has value on its own, but there are also a lot of edge cases that might be easier to test in isolation. I started doing something like this for https://github.com/zenefits/ember-component-apply and something similar and for https://github.com/toranb/ember-cli-hot-loader/pull/25. While both of those approaches work as a "transparent proxy" for components, I don't think they're applicable for ember-tooltips since we rely in one case on transpiling templates on the fly and in the other one to rewrite the AST at buildtime. I only mention those two, because they already cover some of the edge cases for "transparent proxies" and you can look at some of the tests cases for them.

MiguelMadero: Ok, so this seems like it's not an issue. @andrew-zenefits pointed me to https://github.com/sir-dunxalot/ember-tooltips/pull/104/files#diff-8096b1ffac251e79ea18b7393d9f241fR68 which re-computes the `passedPropertiesObject` when any `PASSABLE_OPTIONS` change.

dodeja:
@a15n you might be already aware of this issue but I want to re-surface it. https://github.com/sir-dunxalot/ember-tooltips/issues/107#issuecomment-261391893

Since IE10 is still a widely used browser I think it would be good to support proper rendering. I will test the issue in other IE browsers and report back.

sir-dunxalot:
Ok, to confirm: #107 is a breaking change and should be released as a `2.x.x` patch.

dodeja:
Visibility: hidden did resolve my issue. Only tested IE 10 though.

PR #110, opened 2016-11-07
A15n/support tether components

a15n:
There are situations in which someone would need to extend `tether-tooltip-on-element`, etc instead of `tooltip-on-element`, etc. These situations included...

* extending the `tether-tooltip-on-element`, etc component and to use properties not compatible with a tagless component (which `tooltip-on-element` is)
  * class
  * classNames
  * classNameBindings
  * id

The way I structured the components in the 2.4.0 release didn't allow people to extend the underlying `tether-<x>-on-<y>` components. This fixes that with...

* a `_shouldTargetGrandparentView` which is true in the `<x>-on-component` instances. (https://github.com/sir-dunxalot/ember-tooltips/pull/110/files#diff-14324d21185a9fb331ec5b9c0c7e191cR8)
* explicitly adding the `action.hide` hash to the `tether-popover-on-<y>` instances (https://github.com/sir-dunxalot/ember-tooltips/pull/110/files#diff-665a38a8ad19d6716be00ead6a6446c7R1)
* don't assume the parentView always exists, because it won't when calling `tether-popover-on-<y>` by itself (https://github.com/sir-dunxalot/ember-tooltips/pull/110/files#diff-676a2d0e7fd0e472b2a8af1d3777e951R24)
* adding tests for `tether-<x>-on-<y>` components (no new test failures in beta or canary)


a15n:
@MiguelMadero, @sir-dunxalot FYI

Duncan can you review?

Miguel can you think of any other breaking changes I may have inadvertently introduced with the 2.4.0 release?

Thanks guys

MiguelMadero:
I think this looks good from what I remember. I only quickly scanned #104, so I'll have a deeper look now and see if there is anything else I can find.

MiguelMadero:
I had a look at #104, all good.

a15n:
Great. Can you release this as 2.4.1? Message me if you'd like me to write some release notes

sir-dunxalot:
Done! Release notes would be great @a15n.

Issue #102, opened 2016-10-17
Add test helpers for tooltip assertions


sir-dunxalot:
Issue #101 introduces the idea of public test helpers being beneficial if they can be used in consuming app's own test suites to test tooltips added by ember-tooltips.

We should implement publicly-available assertions like those currently used by this addon's test suite in [test/helpers/assert-visbility.js](https://github.com/sir-dunxalot/ember-tooltips/blob/master/tests/helpers/sync/assert-visibility.js). At minimum, this should include:
- Asserting tooltips are shown
- Asserting tooltips are hidden
- [maybe] Asserting a tooltip has certain content

sir-dunxalot:
Fixed by #114

Issue #74, opened 2016-06-01
Test helpers should be publically available and documented


sir-dunxalot:
Test helpers in some component integration tests should be exported from modules so consuming app developers can use the tooltip test helpers in their own apps.
- assertPosition
- assertVisibility
- Other options?

PR #114, opened 2016-11-22
Public test helpers

a15n:
* exposes public test helpers and [adds documentation](https://github.com/sir-dunxalot/ember-tooltips/pull/114/files#diff-04c6e90faac2675aa89e2176d2eec7d8R414)
  * assertTooltipNotRendered, assertTooltipRendered, assertTooltipVisible, assertTooltipNotVisible, and triggerTooltipTargetEvent
* adds new tests
* fixes tests in every scenario
* removes double-triggered, lazy-render events ([code here](https://github.com/sir-dunxalot/ember-tooltips/pull/114/files#diff-8096b1ffac251e79ea18b7393d9f241fR175) and [tests here](https://github.com/sir-dunxalot/ember-tooltips/pull/114/files#diff-d320863fe14368780962f48a86ff2044R62))
* fixes rapid mouseenter/mouseleave tooltips sticking when `enableLazyRendering=true` ([code here](https://github.com/sir-dunxalot/ember-tooltips/pull/114/files#diff-8096b1ffac251e79ea18b7393d9f241fR146) and [tests here](https://github.com/sir-dunxalot/ember-tooltips/pull/114/files#diff-d320863fe14368780962f48a86ff2044R94). see gif)
* solidifies popover "isElementIn_____" logic and add tests ([here](https://github.com/sir-dunxalot/ember-tooltips/pull/114/files#diff-676a2d0e7fd0e472b2a8af1d3777e951R7))
* remove popover `mousedown` handling in favor of `click` handling ([here](https://github.com/sir-dunxalot/ember-tooltips/pull/114/files#diff-676a2d0e7fd0e472b2a8af1d3777e951R110))
* makes aria-hidden and data-tether-enabled synchronous so that tooltips and popovers can be tested synchronously ([here](https://github.com/sir-dunxalot/ember-tooltips/pull/114/files#diff-aab8720cdae20e1f53c656a7aa537184R109))

TODO

- [x] merge with master and make sure current new tests use new methods
- [x] add documentation for test helpers

### Tooltips sticking before fix
tooltips stick

### Tooltips NOT sticking after fix
tooltips don t stick


sir-dunxalot:
Fixes #102.

sir-dunxalot:
We should make sure this PR addresses #74.

bower.json
```diff
@@ -0,0 +1,12 @@
+{
```
a15n: delete this

a15n:
@sir-dunxalot please review. @JoshSmith this will add test helpers that consuming apps can use.

PS: all test scenarios now pass consistently

screen shot 2016-12-10 at 1 52 44 pm

JoshSmith:
Wow @a15n really impressively done. This is super exciting.

test-support/helpers/ember-tooltips.js
```diff
@@ -0,0 +1,90 @@
...
+
+  return $tooltip;
+}
+
+export function triggerTooltipEvent($element, type, options={}) {
+
+  // TODO why do we allow focusin? why not just focus?
+  const approvedEventTypes = ['mouseenter', 'mouseleave', 'click', 'focus', 'focusin', 'blur'];
+  if (approvedEventTypes.indexOf(type) == -1) {
+    throw Error(`only approvedEventTypes will affect the tooltip target. You used ${type}`);
```
sir-dunxalot: 1) Should this be a warning? 2) Should we echo out the `approvedEventTypes` or provide a link for further information? Maybe this should be included in the documentation.

test-support/helpers/ember-tooltips.js
```diff
@@ -0,0 +1,90 @@
...
+      $element.trigger(type);
+    }
+
+    wasEventTriggered = true;
+  });
+
+  return wasEventTriggered;
+}
+
+export function assertTooltipNotRendered($body, assert, options={}) {
```
sir-dunxalot: I don't remember seeing the `options` param in the documentation. I suggest adding something on what options can be passed for each test helper. Maybe examples of exactly how to use the test helpers would convey this well?

test-support/helpers/ember-tooltips.js
```diff
@@ -0,0 +1,90 @@
...
+  const isTooltipInvisible = $tooltip.attr('aria-hidden') == 'true';
+  const isTooltipTetherDisabled = $tooltip.attr('data-tether-enabled') == 'false';
+
+  assert.ok(isTooltipInvisible && isTooltipTetherDisabled,
+      `assertTooltipNotVisible(): the ember-tooltip should invisible and the tether should be disabled:
+        isTooltipInvisible -> ${isTooltipInvisible} ;
+        isTooltipTetherDisabled -> ${isTooltipTetherDisabled}`);
+}
+
+export function assertTooltipVisible($body, assert, options={}) {
```
sir-dunxalot: Since `assertTooltipNotVisible` and `assertTooltipVisible` share a lot of logic, would is be better to use a private function that houses the logic for these checks given `true` or `false` being passed in as the param?

For example:

```js
function assertHelper(shouldBeVisible) {
  const isTooltipInvisible = $tooltip.attr('aria-hidden') == shouldBeVisible ? 'false' : 'true';

  // etc...
}

export function assertTooltipNotVisible() {
  return assertHelper(false);  
}

export function assertTooltipVisible() {
  return assertHelper(true);
}
```

tests/acceptance/acceptance-test.js
```diff
@@ -0,0 +1,174 @@
+import Ember from 'ember';
```
sir-dunxalot: What coverage is this providing that our existing tests don't? Is this basically testing the test helpers?

tests/integration/components/helpers/trigger-tooltip-event-test.js
```diff
@@ -0,0 +1,21 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import { triggerTooltipEvent } from '../../../helpers/ember-tooltips';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('tooltip-on-element', 'Integration | Helpers | triggerTooltipEvent', {
+  integration: true,
+});
+
+test('triggerTooltipEvent throws appropriate error', function(assert) {
```
sir-dunxalot: Should we test for the events actually being emitted here? I have experienced the brittle nature of events in Ember test before so I'm ok if we don't do that - am just suggesting it.

sir-dunxalot:
Overall, this is a fantastic PR. Great work, again!

These test helpers will really help consuming apps and will further solidify this library as the reliable, robust way for Ember devs to add tooltips to apps in a plug-and-play way.

My comments are mostly around documentation and code comments. The more documentation here the better, IMO. I would like to see a few examples of exactly how devs can use these test helpers. I could see some devs trying to use these in unit tests by accident, for example. We should let them know when they will help and when they won't. We should also make sure the `options` test helper param is documented - I made a review comment for that specific change.

With that said, The logic looked sound. I didn't scan through every single test file because there were so many changed just with the helper reference. however, I read through every test file that had a substantial number of changes.

sir-dunxalot:
As for the future of test helpers, I think we should not make the scope of this PR any bigger but, in the future, it would be good to see helpers for asserting tooltip content and tooltip positions.

JoshSmith:
@sir-dunxalot that covers the most important bases for me.

sir-dunxalot:
One more thought - I assume we are ok with the naming convention of `assertTooltipWhatever` applying to both tooltips and popovers? It seems a bit unnecessary to have two functions that do the same thing but maybe there is a more encompassing name. I'm fine with the current format because `tooltipOrPopover` tends to get quite cumbersome but I just want to bring up the question incase we have better ideas.

README.md
```diff
@@ -410,6 +411,22 @@ Four actions are available for you to hook onto through the tooltip/popover life
 }}
 ```
 
+## Test Helpers
+
+This addon exposes testing helpers which can be used inside of the consuming app's acceptance and integration tests. We use a tooltip-centric naming convention but these can also be used to test popovers.
+
+* `assertTooltipVisible($body, assert)`: asserts if the tooltip is visible. It checks two attributes: aria-hidden and data-tether-enabled.
```
a15n: correct. `assertTooltipVisible` only checks if the tooltip is shown.

README.md
```diff
@@ -410,6 +411,22 @@ Four actions are available for you to hook onto through the tooltip/popover life
...
+* `assertTooltipNotVisible($body, assert)`: asserts if the tooltip is not visible. It checks two attributes: aria-hidden and data-tether-enabled.
+* `assertTooltipRendered($body, assert)`: asserts if the tooltip has been rendered. When enableLazyRendering is true the tooltip will only be rendered after the user has interacted with the $target element.
```
a15n: I'll update the documentation. This is different because a tooltip can be rendered but not visible.

README.md
```diff
@@ -410,6 +411,22 @@ Four actions are available for you to hook onto through the tooltip/popover life
...
+* `assertTooltipNotRendered($body, assert)`: asserts if the tooltip has not been rendered. When enableLazyRendering is true the tooltip will only be rendered after the user has interacted with the $target element.
+* `triggerTooltipEvent($targetElement, 'click')`: triggers an event on the passed element. The event will be triggered within an Ember.run so that the tooltip's asynchronicity is accounted for.
```
a15n: you're right. That would be better. Changing....

test-support/helpers/ember-tooltips.js
a15n: I don't think enough is common to constrict this to a boolean driven helper

tests/acceptance/acceptance-test.js
a15n: Exactly, testing our test helpers

addon/components/lazy-render-wrapper.js
```diff
@@ -133,26 +143,40 @@ export default Ember.Component.extend({
...
+					this.set('_shouldShowOnRender', true);
 				}
 			});
 		});
 	},
 
+	childView: null, // this is set during the childView's didRender and is needed for the hide action
+	actions: {
+		hide() {
+			const childView = this.get('childView');
```
a15n: addressed

addon/components/lazy-render-wrapper.js
```diff
@@ -121,6 +117,20 @@ export default Ember.Component.extend({
 
 		return false;
 	}),
+	_shouldShowOnRender: false,
+
+	event: 'hover', // hover, click, focus, none
+	entryInteractionEvents: computed('event', function() {
```
a15n: done

a15n:
@sir-dunxalot thanks a lot for your feedback. I've addressed all of it and made changes in these two commits. I believe this is ready to merge/release now. I'll prepare some release notes.

angelvivienne:
awesome work here, become a fan for this!

The test helpers implemented so far are really helpful, can we also add one as checking the content of tooltip? like getting the content of the text-field by `.text()`

sir-dunxalot:
@angelvivienne I just opened #128 to track this request. I agree!

sir-dunxalot:
Fixes #118

JoshSmith:
🙌  everyone, especially @a15n!

PR #117, opened 2016-11-23
Remove some-component from addon files

sir-dunxalot:
This PR removes the `some-component` component from addon files. We only need it in tests and it already exists for tests [here](https://github.com/sir-dunxalot/ember-tooltips/blob/547ec11ccb0725cbfe92234445cd8f128e43135d/tests/dummy/app/components/some-component.js). If we don't remove these files, every consuming app will have a `some-component` component added to it by this addon.


a15n:
LGTM

PR #119, opened 2016-11-28
allow for classNames to be passed in

### LINKED ISSUE #112: passing classNames to tooltip-on-element doesn't work anymore
Just upgraded to 2.4.0 and now passing `classNames` into `tooltip-on-element` doesn't add that class to the div.

```hbs
{{#tooltip-on-element
   side="right"
   tooltipIsVisible=true
  event="none"
  classNames="got-it"
}}
   Some text
{{/tooltip-on-element}}
```

SirZach:
closes #112

### DIFF
--- a/tests/integration/components/tooltip-on-element-test.js
+++ b/tests/integration/components/tooltip-on-element-test.js
@@ -40,3 +40,18 @@ test('it has the proper aria-describedby tag', function(assert) {
   assert.equal(this.$(`#${describedBy}`).text().trim(), 'Some info in a tooltip.');
   assert.equal(describedBy.indexOf('#'), '-1');
 });
+
+test('it renders passed in classes', function(assert) {
+  this.render(hbs`
+    <div class="target">
+      {{#tooltip-on-element classNames='got-it another-class'}}
+        Got it
+      {{/tooltip-on-element}}
+    </div>
+  `);
+
+  const $tooltip = this.$().find('.ember-tooltip');
+
+  assert.ok($tooltip.hasClass('got-it'), 'tooltip has the got-it class');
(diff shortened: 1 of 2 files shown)


tests/integration/components/tooltip-on-element-test.js
```diff
@@ -40,3 +40,18 @@ test('it has the proper aria-describedby tag', function(assert) {
   assert.equal(this.$(`#${describedBy}`).text().trim(), 'Some info in a tooltip.');
   assert.equal(describedBy.indexOf('#'), '-1');
 });
+
+test('it renders passed in classes', function(assert) {
```
a15n: this is the exact type of test that we need, but can you add it to the `pass-through-test.js` file instead? You can just add classNames to the existing test and add more `.hasClass` calls

a15n:
This fixes the issue with `classNames`. thank you SirZach. While we're at it can you do a bit of research and see if there are any other attributes that need to be passed through? I think attributeBindings would be another that we should add coverage for during the same PR.

SirZach:
@a15n you cannot pass in `attributeBindings`. The component itself needs to have the `attributeBindings` setup. I can take a crack with a separate PR to add all the attributes

sir-dunxalot:
@SirZach Can you merge in master and update the test suite? Then we should be good to merge and release this improvement!

a15n:
thanks for finding this @SirZach. We've actually addressed the issue with this PR. 

https://github.com/sir-dunxalot/ember-tooltips/pull/133

SirZach:
@a15n sorry for my absence and thanks for following it through.

PR #120, opened 2016-11-29
[feature] `disabled` attribute

knownasilya:
Allows a `disabled` attribute to prevent showing the tooltip by any of the available methods:

```hbs
{{#tooltip-on-component disabled=true}}
  My Tooltip
{{/tooltip-on-component}}
```

Talked about this with @sir-dunxalot on Slack.

### DIFF
--- a/addon/components/lazy-render-wrapper.js
+++ b/addon/components/lazy-render-wrapper.js
@@ -94,8 +94,8 @@ export default Ember.Component.extend({
 	enableLazyRendering: false,
 	_hasUserInteracted: false,
 	_hasRendered: false,
-	_shouldRender: computed('isShown', 'disabled', 'tooltipIsVisible', 'enableLazyRendering', '_hasUserInteracted', function() {
-		// if isShown, tooltipIsVisible, !enableLazyRendering, or _hasUserInteracted then
+	_shouldRender: computed('disabled', 'isShown', 'tooltipIsVisible', 'enableLazyRendering', '_hasUserInteracted', function() {
+		// if not disabled, isShown, tooltipIsVisible, !enableLazyRendering, or _hasUserInteracted then
 		// we return true and set _hasRendered to true because
 		// there is never a scenario where this wrapper should destroy the tooltip
 
--- a/addon/components/lazy-render-wrapper.js
+++ b/addon/components/lazy-render-wrapper.js
@@ -94,7 +94,7 @@ export default Ember.Component.extend({
 	enableLazyRendering: false,
 	_hasUserInteracted: false,
 	_hasRendered: false,
-	_shouldRender: computed('isShown', 'tooltipIsVisible', 'enableLazyRendering', '_hasUserInteracted', function() {
+	_shouldRender: computed('isShown', 'disabled', 'tooltipIsVisible', 'enableLazyRendering', '_hasUserInteracted', function() {
 		// if isShown, tooltipIsVisible, !enableLazyRendering, or _hasUserInteracted then
 		// we return true and set _hasRendered to true because
 		// there is never a scenario where this wrapper should destroy the tooltip
@@ -103,7 +103,7 @@ export default Ember.Component.extend({
 
 			return true;
 
-		} else if (this.get('isShown') || this.get('tooltipIsVisible')) {
+		} else if (!this.get('disabled') && (this.get('isShown') || this.get('tooltipIsVisible'))) {
 
 			this.set('_hasRendered', true);
 			return true;
--- a/addon/components/lazy-render-wrapper.js
+++ b/addon/components/lazy-render-wrapper.js
@@ -32,6 +32,7 @@ const PASSABLE_PROPERTIES = [
 	'keepInWindow',
 	'side',
 	'showOn',
+  'disabled',
 	'spacing',
 	'isShown',
 	'tooltipIsVisible',
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,13 @@ sudo: false
 
 cache:
   directories:
-    - node_modules
+    - $HOME/.npm
+    - $HOME/.cache # includes bowers cache
 
 env:
-  - EMBER_TRY_SCENARIO=default
-  - EMBER_TRY_SCENARIO=ember-1.13
+  # we recommend testing LTS's and latest stable release (bonus points to beta/canary)
+  - EMBER_TRY_SCENARIO=ember-lts-2.4
+  - EMBER_TRY_SCENARIO=ember-lts-2.8
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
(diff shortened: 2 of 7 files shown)


knownasilya:
Only thing I'm not sure about is not showing the tooltip if `disabled=true isShown=true`..

knownasilya:
Tests are failing for unrelated reasons.

knownasilya:
So I added LTS targets and those all pass along with 1.13, but delay is still broken with 2.10. This feature should still be mergeable, since this test failure is unrelated.

sir-dunxalot:
@a15n I don't want to merge this without a review from you given to the effects of this on `isShown`. I will also do a review.

addon/components/lazy-render-wrapper.js
```diff
@@ -93,7 +94,7 @@ export default Ember.Component.extend({
 	enableLazyRendering: false,
 	_hasUserInteracted: false,
 	_hasRendered: false,
-	_shouldRender: computed('isShown', 'tooltipIsVisible', 'enableLazyRendering', '_hasUserInteracted', function() {
+	_shouldRender: computed('isShown', 'disabled', 'tooltipIsVisible', 'enableLazyRendering', '_hasUserInteracted', function() {
```
a15n: let's move `disabled` to be the first property. It is the most important.

addon/components/lazy-render-wrapper.js
```diff
@@ -93,7 +94,7 @@ export default Ember.Component.extend({
...
 		// if isShown, tooltipIsVisible, !enableLazyRendering, or _hasUserInteracted then
```
a15n: can you add `disabled` to this note as well

addon/components/lazy-render-wrapper.js
```diff
@@ -102,7 +103,7 @@ export default Ember.Component.extend({
 
 			return true;
 
-		} else if (this.get('isShown') || this.get('tooltipIsVisible')) {
+		} else if (!this.get('disabled') && (this.get('isShown') || this.get('tooltipIsVisible'))) {
 
 			this.set('_hasRendered', true);
 			return true;
```
a15n: I can't comment below this line so see lines....

111: what if `enableLazyRendering=false` and `disabled=true`?

129:  what if `enableLazyRendering=false` but `_shouldRender=false`? The didInsertElement will still apply $parent event handling and could still render the tooltip, even if it's never shown (not clean)

knownasilya: should I just handle `disabled` once before all of these conditionals, and `return false`?

knownasilya: And what about in 128, if i skip the parent event stuff, and then disabled gets set to false, would that be rerun anywhere else?

a15n:
So the `disabled` property will prevent the tooltip from being rendered if used on the `tooltip-on-element` and prevent the tooltip from being shown on `tether-tooltip-on-element`. It will prevent it regardless of `isShown` and user interaction. Am I understanding this correctly?

You can achieve the "no show" behavior if you set `isShown=false` and `event="none"` or by wrapping your tooltip in an `{{#if}}` block. @jliuzen how would you do this if you wanted to disable `z-help-text-popover` for mobile users? Assuming you had a boolean for `isMobile`.

I'm a little hesitant to introduce this because there are existing ways to get the same behavior and with more options comes more complexity. From a technical point of view this PR looks good though. Can you explain a little more about why a `disabled` property is the best option?

a15n:
@aaminev would this be useful in the scenario when there is no tooltip text? You could do something like...

```
{{#tooltip-on-element disabled=(not tooltipText.length)}}
  {{tooltipText}}
{{/tooltip-on-element}}
```

a15n:
@knownasilya also thanks for your work to add LTS support! I definitely think that's a great idea but I think it should be done in a separate PR. And also I'd prefer that we keep master green so we'd need to add 2.1.0 to the list of allow_failures until the delay test is fixed.

addon/components/lazy-render-wrapper.js
a15n: > "should I just handle disabled once before all of these conditionals, and return false?"

I think if you add this it'll take care of _shouldRender
```
if (this.get('_hasRendered')) {
  return true; // this always needs to return true if it has ever been rendered
} else if (isDisabled) {
  return false;
....
```

a15n: > "And what about in 128, if i skip the parent event stuff, and then disabled gets set to false, would that be rerun anywhere else?"

You're right. If we people change isDisabled we would need to add the event listeners to the $parent. I think maybe we should make the $parent event handling a function that can be called on didInsertElement and by observing `isDisabled`. 

We'd need good tests to cover this. It's getting fairly complex

knownasilya:
Maybe this is too big of a feature for something that could be replaced by a wrapper.. e.g.

```hbs
{{desktop-tooltip}}
  My Text
{{/desktop-tooltip}}
```

which just wraps the tooltip in an `{{#if`.

a15n:
Agreed, I think the ideal way would be to create you're own wrapper.

sir-dunxalot:
Another solution could be to set `event='none'` in scenarios where you don't want to render the tooltip.

knownasilya:
@sir-dunxalot so change it to `none` and then change it back when you want it to show? Becuase wrapping it in a component doesn't really work I get no behavior unless i set `tagName: ''` and then I get errors: 

```
Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at Object.clear (http://stocktonca.mg.dev/assets/vendor.js:59330:20)
    at UpdatableBlockTracker.reset (http://stocktonca.mg.dev/assets/vendor.js:59667:56)
    at TryOpcode.handleException (http://stocktonca.mg.dev/assets/vendor.js:68468:113)
    at UpdatingVMFrame.handleException (http://stocktonca.mg.dev/assets/vendor.js:68650:35)
    at UpdatingVM._throw [as throw] (http://stocktonca.mg.dev/assets/vendor.js:68363:37)
    at Assert.evaluate (http://stocktonca.mg.dev/assets/vendor.js:63745:25)
    at UpdatingVM.execute (http://stocktonca.mg.dev/assets/vendor.js:68350:24)
    at RenderResult.rerender (http://stocktonca.mg.dev/assets/vendor.js:68289:16)
    at RootState._this.render (http://stocktonca.mg.dev/assets/vendor.js:25879:18)
    at runInTransaction (http://stocktonca.mg.dev/assets/vendor.js:36838:28)
```

knownasilya:
Related issue: https://github.com/yapplabs/ember-tether/issues/35 so this doesn't work with 2.10 if you put it in an `{{#if`.

knownasilya:
So the workaround for that would be to wrap `{{#tooltip-on-element` with a `<div>`, but this doesn't work for me because then the tooltip doesn't show since it's on the wrong element.

SirZach:
@knownasilya was this thread continued anywhere else or is it not addressed? I see https://github.com/yapplabs/ember-tether/issues/35 is still open which is blocking the upgrade to 2.11 at the moment

knownasilya:
It hasn't continues. Basically I did a wrapper component like so:

```hbs
{{#unless media.isMobile}}
  <div>
    {{#tooltip-on-element side=side delay=delay duration=duration}}
      {{yield}}
    {{/tooltip-on-element}}
  </div>
{{/unless}}
```

SirZach:
@knownasilya did you have a solution for `{{tooltip-on-component}}`?

knownasilya:
I didn't have that requirement. I think this PR would have to be completed to support that.

PR #124, opened 2016-12-09
BUGFIX checkif component is destroyed before trying to set properties…

bargar:
… on it

This is something that came up for me when I upgraded an app to Ember 2.10 with the Glimmer2 rendering engine. It is possible in some scenarios in my component to remove a tooltip so quickly that this `stopTether` runs after it is already destroyed. A simple `isDestroyed` check avoids the bug.

Thank you for your consideration...

### DIFF
--- a/addon/components/tether-tooltip-and-popover.js
+++ b/addon/components/tether-tooltip-and-popover.js
@@ -482,10 +482,13 @@ export default EmberTetherComponent.extend({
 
   stopTether() {
     run.schedule('afterRender', () => {
-      // can't depend on `_tether.enabled` because it's not an
-      // Ember property (so won't trigger cp update when changed)
-      this.set('_isTetherEnabled', false);
-      this.get('_tether').disable();
+      if (!this.isDestroyed && !this.isDestroying) {
+
+        // can't depend on `_tether.enabled` because it's not an
+        // Ember property (so won't trigger cp update when changed)
+        this.set('_isTetherEnabled', false);
+        this.get('_tether').disable();
+      }
     });
   },
 


addon/components/tether-tooltip-and-popover.js
```diff
@@ -454,10 +454,13 @@ export default EmberTetherComponent.extend({
 
   stopTether() {
     run.schedule('afterRender', () => {
-      // can't depend on `_tether.enabled` because it's not an
-      // Ember property (so won't trigger cp update when changed)
-      this.set('_isTetherEnabled', false);
-      this.get('_tether').disable();
+      if (!this.isDestroyed) {
```
sir-dunxalot: Should we check for `this.isDestroying` instead? I say that because `isDestroying` is set to `true` even earlier than `this.isDestroyed` so it may be more robust.

Also, are you not using `get` because there is no need for a destroyed component? I ask just because I'm not sure myself.

sir-dunxalot:
The test suit is currently failing. A quick solution to this may be waiting until #114 is merged because that has an all-green test suite even for canary and beta builds.

sir-dunxalot:
@bargar If you want to merge this in, I suggest rebasing with master or (if it's simple) just closing this PR and opening another PR off of the current state of master.

bargar:
Will rebase, thanks for the heads up!

addon/components/tether-tooltip-and-popover.js
bargar: Happy to include `isDestroying` as well. Just because I've not seen that error state doesn't mean we should be too sure it couldn't happen.

I've updated the check to be like this example from Ember itself:

https://github.com/emberjs/ember.js/blob/d5acb2d4ec989aa33959cd31cddcd1c854f9919f/packages/ember-runtime/lib/mixins/promise_proxy.js#L22

This and other similar examples are what I'm following when I directly access the property rather than through `get`.

bargar:
@sir-dunxalot rebased and updated to include `isDestroying`.

Let me know if you'd like further tweaks and I'll make 'em.

sir-dunxalot:
This looks good! It will be merged and released in the near future.

sir-dunxalot:
Thank you @bargar! This is another good step towards making this library more robust.

sir-dunxalot:
Released as 2.6.1 ([release notes here](https://github.com/sir-dunxalot/ember-tooltips/releases/tag/2.6.1)).

bargar:
@sir-dunxalot looking good, switched back to official repo now that this is incorporated. Thanks for incorporating!

Issue #109, opened 2016-11-05
3.0.0 release


sir-dunxalot:
Changes that need to be made for the `3.0.0` release:

- [x] Deprecate all properties that throw a deprecation warning in the 2.3.2 (#125)
- [x] Set `enableLazyRendering` to true by default (#125)
- [x] Remove `<x>-on-component` and `<x>-on-element` in favor of `ember-<x>` and an `isTargetComponent` property
- [x] Add publicly available test helpers for consuming apps (#114)
- [x] Deprecate `delay` in favor of `showDelay` (#125)
- [x] Fix IE multi line bug (#137)
- [x] Add code linting (#154)
- [x] Fix #152 by making `ember-hash-helper-polyfill` a required `peerDependency` for apps using Ember less than `2.3.0`.

sir-dunxalot:
@a15n Anything else we want to add to this list? I created a `3.0.0 milestone` to group anything else relevant to this release.

a15n:
I think that's it for now. Thanks for making this ticket

a15n:
Maybe remove CSS styles from the tooltip. I had expected the popovers to be completely unstyled. At least in my app I'm having to override a lot of these styles...

https://github.com/sir-dunxalot/ember-tooltips/commit/bc55092cc30de896c0ad7ea13a02e1b60e6b8ae3#commitcomment-19321234

sir-dunxalot:
So that's an interesting discussion. For the majority of developers, I think this should be a plug-and-play solution where tooltips will work with an install but you can customize them with additional work. Maybe there is an alternative where an option can be set that would exclude the styles from being concatenated into the consuming app's `vendor.css` file. This option could be set in the `ember-cli-build` file.

a15n:
You bring up a good point with the plug-and-play majority needs. I think since that's the case the current approach is good.

sir-dunxalot:
Ok, sounds good. Happy to explore the option of excluding stylesheets if that would alleviate the pain for those who do overwrite tooltip and popover styles.

sir-dunxalot:
@a15n Thoughts on adding publicly available test helpers for consuming apps that test tooltips?

a15n:
Good idea. Can you message me with how to best do this?

sir-dunxalot:
I've done the first couple of steps on [this branch](https://github.com/sir-dunxalot/ember-tooltips/tree/public-test-helpers). The new file is in the `/test-support` directory.

[...]

PR #125, opened 2016-12-11
Deprecate properties for 3.0

sir-dunxalot:
~~**Prerequisite: #114**~~

This PR removes the following deprecated properties:

- `onTooltipDestroy` (now `onDestroy`)
- `onTooltipHide` (now `onHide`)
- `onTooltipRender` (now `onRender`)
- `onTooltipShow` (now `onShow`)
- `tooltipIsVisible` (now `isShown`)
- `delay` (now `showDelay`)

In addition, the following breaking changes have been made:

- `enableLazyRendering` is `true` by default (previously it was `false`)


sir-dunxalot:
@a15n I have merged your `public-test-helpers` branch into this so that once #114 is merged this will be a very seamless merge and the test suite will be passing. However, there are three tests failing and all are to do with the same thing: asserting that a tooltip is hidden when certain event options are passed. You can [see the failing tests here](https://travis-ci.org/sir-dunxalot/ember-tooltips/jobs/182963762). I think this may actually be a bug from earlier code. Now `enableLazyRendering` is `true` by default, when any user interaction renders the tooltip, it would seem that the tooltips are being made visible (at least somewhat). For example, when the `click` event is triggered on a tooltip target, the tooltip's `aria-hidden` attribute will be `false` but its opacity will still be 0. As such, the test helpers introduced in #114 fail the test given the assertions in the helpers. You can recreate the issue simply by pulling down the `deprecate-properties-for-3.0` branch and running the test suite.

I can get around this issue by setting `enableLazyRendering=false` in the test templates. However, I have tried to make the tests as 'authentic' as possible knowing most developers won't be setting `enableLazyRendering=false`. What do you think the best course of action is here? Are you aware of a bug or even functionality that might explain what is happening here?

sir-dunxalot:
@a15n Tests are failing but this is ready for a code review by yourself. Also, see the question above.

sir-dunxalot:
This PR knocks out 3 of the 5 todos in #109

sir-dunxalot:
Closing given the large changes that happened on master branch since this PR.

Issue #129, opened 2016-12-15
ember-tooltips is broken in ember 2.10.1


hoboman313:
Everything works fine with ember 2.10.0. Upgrade to 2.10.1 and I get
```
Assertion Failed: An action named 'hide' was not found in <...@component:tooltip-on-element::ember12751>
```

andrew-zenefits:
which version of ember-tooltips are you using? I think I've fixed this in 2.5.0

hoboman313:
I'm on 2.5.0

andrew-zenefits:
mmmmm tests are passing up through ember-canary https://travis-ci.org/sir-dunxalot/ember-tooltips

We need to add specific LTS support but it seems like this should be working through canary. Could you try `nombom`? The tooltip-on-element component definitely has a hide action

tooltip-on-element extends lazy-render-wrapper

https://github.com/sir-dunxalot/ember-tooltips/blob/master/addon/components/lazy-render-wrapper.js

hoboman313:
@andrew-zenefits My mistake. Forgot that yarn shrinkwraps. I was not on 2.5.0 and it looks like the issue is fixed now.

PR #132, opened 2016-12-15
add LTS 2.4 and 2.8 support

a15n:
* adds LTS support for version 2.4 and 2.8
* removes the `default` scenario.
  * the `default` scenario was the same as the `ember-release` scenario because our ember-version is ^2.6.0, which currently resolves to 2.10.1

Thanks for the inspiration @knownasilya https://github.com/sir-dunxalot/ember-tooltips/pull/120

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,15 @@ sudo: false
 
 cache:
   directories:
-    - node_modules
+    - $HOME/.npm
+    - $HOME/.cache # includes bower's cache
 
 env:
+  # we recommend testing LTS's and latest stable release (bonus points to beta/canary)
   - EMBER_TRY_SCENARIO=default
   - EMBER_TRY_SCENARIO=ember-1.13
+  - EMBER_TRY_SCENARIO=ember-lts-2.4
+  - EMBER_TRY_SCENARIO=ember-lts-2.8
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
@@ -22,7 +26,9 @@ matrix:
 before_install:
   - npm config set spin false
   - npm install -g bower
+  - bower --version
   - npm install phantomjs-prebuilt
+  - node_modules/phantomjs-prebuilt/bin/phantomjs --version
 
 install:
   - npm install
@@ -30,5 +36,5 @@ install:
 
 script:
   # Usually, it's ok to finish the test scenario without reverting
-  #  to the addon's original dependency state, skipping "cleanup".
-  - ember try $EMBER_TRY_SCENARIO test --skip-cleanup
+  # to the addon's original dependency state, skipping "cleanup".
+  - ember try:one $EMBER_TRY_SCENARIO test --skip-cleanup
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,11 +7,15 @@ sudo: false
 
 cache:
   directories:
-    - node_modules
+    - $HOME/.npm
+    - $HOME/.cache # includes bower's cache
 
 env:
+  # we recommend testing LTS's and latest stable release (bonus points to beta/canary)
   - EMBER_TRY_SCENARIO=default
```
a15n: @sir-dunxalot `EMBER_TRY_SCENARIO=default` means it runs whatever is in the bower.json, which is `"ember": "^2.6.0",`

Seeing as this is between lts-2.4 and 2.8 should we move the line there? Or should we update the bower.json to a different version and remove the default scenario?

andrew-zenefits:
tests now take ~10 minutes to run. :-/

travis is only running 5 tests at a time but there are 7 scenarios in our test suite. Maybe we can bump up the version in travis somehow...

a15n:
tests now take ~10 minutes to run. :-/

travis is only running 5 tests at a time but there are 7 scenarios in our test suite. Maybe we can bump up the version in travis somehow...

.travis.yml
sir-dunxalot: I think we should bump up the `bower.json` by updating Ember CLI as a whole to the latest version.

sir-dunxalot:
Also would like to see the test suite speed not take 11 minutes if there is a way to make them run faster. Maybe there is a way to raise the 5-test limit.

a15n:
Just did some research and based on what I can find we're limited to 5 builds in a free plan

https://twitter.com/travisci/status/651856122559774722

.travis.yml
a15n: so if I bump to the latest version then we won't need a default build, because that'll be covered in the `ember-release` scenario. Is that correct? If so that'll put us to 6 scenarios we're testing.

sir-dunxalot: Yes, I believe that is the case. I would be nice to keep it to five builds per our conversations about test suite speed. If we do sacrifice one for now, maybe it's `ember-canary`?

a15n: Personally I think that it'd be better to optimize for comprehensive test coverage rather than test run time in the CI environment. Within the next 6 months we'll have another LTS version to support and I don't think we would want to make a sacrifice then as well to try to maintain a 5-test maximum limit.

Also, the fact that we have canary test coverage has saved me from spending time debugging when unnecessary, so it seems like a time-saver to me as a maintainer.

https://github.com/sir-dunxalot/ember-tooltips/issues/129

a15n: so I studied semver a little more and a version of `^2.6.0` include everything greater than a particular version in the same major range. So this resolves to `Ember 2.10.1`, which means we can remove the `default` scenario because it is the same as the `release` scenario.

a15n:
This conversation is hidden but very relevant to this PR

https://github.com/sir-dunxalot/ember-tooltips/pull/132#discussion_r92633037

@sir-dunxalot I feel this is ready to merge

PR #137, opened 2016-12-19
add `white-space: normal;` because IE doesnt support `initial`

### LINKED ISSUE #126: Multiline tooltip not working in IE11
Oh lovly IE:
tooltips

Everything works fine in Chrome, Edge, etc.

luxferresum:
This fixes #126.
IE doest support `white-space: initial;`, so it will fall back to `inherit`.

### DIFF
--- a/addon/styles/addon.scss
+++ b/addon/styles/addon.scss
@@ -31,7 +31,7 @@
   color: #fff;
   background: #3a3c47;
   text-shadow: -1px -1px 0 rgba(0,0,0,0.2);
-  white-space: normal;
+  white-space: normal; // IE doesnt support initial so fall back to normal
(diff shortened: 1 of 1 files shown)


addon/styles/addon.scss
```diff
@@ -31,6 +31,7 @@
   color: #fff;
   background: #3a3c47;
   text-shadow: -1px -1px 0 rgba(0,0,0,0.2);
+  white-space: normal;
```
sir-dunxalot: Can you add a comment explaining why this line has been added?

sir-dunxalot:
Merged! Thanks for the research and time to debug this. Will release shortly.

PR #145, opened 2016-12-22
Spacing test helper

sir-dunxalot:
This PR adds a new public test helper, `assertTooltipSpacing()`. This helper lets you assert that the tooltip is a given number of pixels from the target on a given side on the target.

This is designed to test the [spacing option](https://github.com/sir-dunxalot/ember-tooltips#spacing).

Usage is relatively simple and matches the implementation of existing test helpers:

```js
assertTooltipSpacing(assert, {
  side: 'right',
  spacing: 10, // Tooltip should 10px from right of target
});
```

You can also pass in selectors:

```js
assertTooltipSpacing(assert, {
  side: 'top',
  spacing: 20, // Tooltip should be 20px above target
  targetSelector: '.test-tooltip-target',
  tooltipSelector: '.test-tooltip',
});
```

This PR refactored the recently added `assertTooltipSide()` test helper to make the code more DRY.

### DIFF
--- a/test-support/helpers/ember-tooltips.js
+++ b/test-support/helpers/ember-tooltips.js
@@ -5,6 +5,46 @@ const { $, run } = Ember;
 const tooltipOrPopoverSelector = '.ember-tooltip, .ember-popover';
 const tooltipOrPopoverTargetSelector = '.ember-tooltip-or-popover-target';
 
+/**
+@method getPositionDifferences
+@param String side The side the tooltip should be on relative to the target
+
+Given a side, which represents the side of the target that
+the tooltip should render on this method identifies whether
+the tooltip or the target should be further away from the
+top left of the window.
+
+For example, if the side is 'top' then the target should
+be further away from the top left of the window than the
+tooltip because the tooltip should render above the target.
+
+If the side is 'right' then the tooltip should be further
+away from the top left of the window than the target
+because the tooltip should render to the right of the
+target.
+
+This method then returns an object with two numbers:
+
+- `greaterNumber` (expected greater number given the side)
+- `lesserNumber` (expected lesser number given the side)
+
+These numbers can be used for calculations like determining
+whether a tooltip is on the correct side of the target or
+determining whether a tooltip is the correct distance from
+the target on the given side.
+*/
+
+function getPositionDifferences(side) {
+  const { targetPosition, tooltipPosition } = getTooltipPosition();
+  const one = targetPosition[side];
+  const two = tooltipPosition[getOppositeSide(side)];
+  const shouldBeBeforeTarget = side === 'top' || side === 'left';
+  const greaterNumber = shouldBeBeforeTarget ? one : two;
+  const lesserNumber = shouldBeBeforeTarget ? two : one;
+
+  return { greaterNumber, lesserNumber };
+}
+
 function getTooltipFromBody(selector=tooltipOrPopoverSelector) {
   // we have to .find() tooltips from $body because sometimes
   // tooltips and popovers are rendered as children of <body>
@@ -33,7 +73,16 @@ function getTooltipTargetFromBody(selector = tooltipOrPopoverTargetSelector) {
   return $tooltipTarget;
 }
 
-function validateSide(side) {
+function getOppositeSide(side) {
+  switch (side) {
+    case 'top': return 'bottom'; break;
+    case 'right': return 'left'; break;
+    case 'bottom': return 'top'; break;
+    case 'left': return 'right'; break;
+  }
+}
+
+function validateSide(side, testHelper = 'assertTooltipSide') {
   const sideIsValid = (
     side === 'top' ||
     side === 'right' ||
@@ -45,7 +94,7 @@ function validateSide(side) {
   use Ember.assert because assert is passed in from QUnit */
 
   if (!sideIsValid) {
-    Ember.assert(`You must pass side like assertTooltipSide(assert, { side: 'top' }); Valid options for side are top, right, bottom, and left.`);
+    Ember.assert(`You must pass side like ${testHelper}(assert, { side: 'top' }); Valid options for side are top, right, bottom, and left.`);
   }
 }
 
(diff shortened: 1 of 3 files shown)


test-support/helpers/ember-tooltips.js
```diff
@@ -5,7 +5,49 @@ const { $, run } = Ember;
...
+  const distanceToTarget = targetPosition[side];
+  const distanceToTooltip = tooltipPosition[getOppositeSide(side)];
+  const shouldTooltipBeCloserThanTarget = side === 'top' || side === 'left';
+  const expectedGreaterDistance = shouldTooltipBeCloserThanTarget ? distanceToTarget : distanceToTooltip;
+  const expectedLesserDistance = shouldTooltipBeCloserThanTarget ? distanceToTooltip : distanceToTarget;
+
+  return { expectedGreaterDistance, expectedLesserDistance };
+}
+
+function getTooltipFromBody(selector=tooltipOrPopoverSelector) {
```
sir-dunxalot: I've used `function whatever` instead of `const whatever = function` before and in this file so I thought it best to standardize our private functions to use `function whatever`.

test-support/helpers/ember-tooltips.js
```diff
@@ -33,6 +75,43 @@ const getTooltipTargetFromBody = function(selector = tooltipOrPopoverTargetSelec
...
+  use Ember.assert because assert is passed in from QUnit */
+
+  if (!sideIsValid) {
+    Ember.assert(`You must pass side like ${testHelper}(assert, { side: 'top' }); Valid options for side are top, right, bottom, and left.`);
+  }
+}
+
+function getTooltipPosition(options = {}) {
+  const $target = getTooltipTargetFromBody(options.targetSelector);
+  const targetPosition = $target[0].getBoundingClientRect();
```
sir-dunxalot: @a15n With reference to your question about why `getBoundingClientRect()` is better than jQuery's `position()` method, using `getBoundingClientRect()` avoids [this mess](https://github.com/sir-dunxalot/ember-tooltips/blob/01991d37bd1ad0c31da89606432e4c8d57a69321/tests/integration/components/spacing-test.js#L11).

test-support/helpers/ember-tooltips.js
```diff
@@ -105,37 +184,39 @@ export function assertTooltipVisible(assert, options={}) {
...
-  const expectedSide = options.side || 'top';
-  const $target = getTooltipTargetFromBody(options.targetSelector);
-  const targetPosition = $target[0].getBoundingClientRect();
-  const $tooltip = getTooltipFromBody(options.selector);
-  const tooltipPosition = $tooltip[0].getBoundingClientRect();
+  assert.ok(expectedGreaterDistance > expectedLesserDistance,
+    `Tooltip should be on the ${side} side of the target`);
+}
+
+export function assertTooltipSpacing(assert, options) {
```
sir-dunxalot: I thought about making `assertTooltipSide()` and `assertTooltipSpacing()` a single helper, given that `assertTooltipSpacing()` requires a `side` but decided to keep one assertion per option. Happy to reconsider this.

a15n: I think this is a good approach

test-support/helpers/ember-tooltips.js
```diff
@@ -5,7 +5,49 @@ const { $, run } = Ember;
 const tooltipOrPopoverSelector = '.ember-tooltip, .ember-popover';
 const tooltipOrPopoverTargetSelector = '.ember-tooltip-or-popover-target';
 
-const getTooltipFromBody = function(selector=tooltipOrPopoverSelector) {
+/**
+@method getPositionDifferences
+@param String side The side the tooltip should be on relative to the target
+
+Given a side, which represents the side of the target that
+the tooltip should render on this method identifies whether
```
a15n: I think you need a second comma...

```
Given a side, which represents the side of the target that the tooltip should render, (comma added here) on this method identifies whether
```

tests/integration/components/spacing-test.js
```diff
@@ -1,49 +1,37 @@
...
 
-  assertSpacing(assert, this, 10);
+  assertTooltipSpacing(assert, {
+    side: 'top',
+    spacing: 10,
+  });
 
 });
 
 test('tooltip-on-element shows with showOn spacing=default', function(assert) {
```
a15n: can you update this test name to include `spacing=20`

tests/integration/components/spacing-test.js
```diff
@@ -1,49 +1,37 @@
...
 
-  assert.expect(1);
+  assert.expect(2);
 
   /* Check custom spacing */
 
   this.render(hbs`{{tooltip-on-element spacing=20}}`);
```
a15n: can we also change this to `side="right"` so we can test both attributes of the helper?

test-support/helpers/ember-tooltips.js
```diff
@@ -105,37 +184,39 @@ export function assertTooltipVisible(assert, options={}) {
...
-    assert.ok(targetPosition.left > tooltipPosition.right,
-      'Tooltip should be left of the target');
+  validateSide(side, 'assertTooltipSpacing');
+
+  if (typeof spacing !== 'number') {
+    Ember.assert(`You must pass spacing as a number like assertTooltipSpacing(assert, { side: 'top', spacing: 10 });`);
   }
+
+  const { expectedGreaterDistance, expectedLesserDistance } = getPositionDifferences(options);
+  const actualSpacing = Math.round(expectedGreaterDistance - expectedLesserDistance);
```
a15n: do you need to allow for a margin of error like before?

https://github.com/sir-dunxalot/ember-tooltips/pull/145/files#diff-b7eb48ff1a009e6b663bff11a464e390L15

test-support/helpers/ember-tooltips.js
```diff
@@ -33,6 +75,43 @@ const getTooltipTargetFromBody = function(selector = tooltipOrPopoverTargetSelec
...
+
+  /* We make sure the side being tested is valid. We
+  use Ember.assert because assert is passed in from QUnit */
+
+  if (!sideIsValid) {
+    Ember.assert(`You must pass side like ${testHelper}(assert, { side: 'top' }); Valid options for side are top, right, bottom, and left.`);
+  }
+}
+
+function getTooltipPosition(options = {}) {
```
a15n: this might be a misnomer because it gets targetPosition and tooltipPosition

a15n:
also, how do you feel about using one assert within the assertTooltipSpacing function? I enforced that with the other helpers. I feel it's a bit odd from a user's point of view for `assertTooltipSpacing` or any related helpers to call assert more than once.

test-support/helpers/ember-tooltips.js
sir-dunxalot: Good point. Will update.

test-support/helpers/ember-tooltips.js
sir-dunxalot: Agreed. Will update.

test-support/helpers/ember-tooltips.js
sir-dunxalot: I was expecting to need to do this but every test passed, including when I messed about with different values for `spacing`. I think the new `getBoundingClientRect()` approach has made the tests more accurate and, thus, there isn't a need for the margin of error.

tests/integration/components/spacing-test.js
sir-dunxalot: Good catch, will do.

sir-dunxalot:
If we wanted to only have one assertion in `assertTooltipSpacing()` would you still test for side and spacing or would you test for spacing and ignore side? If we did the latter, then the spacing could be correct (20px from the target) but the side could be completely wrong. Maybe that's how it should function but I just want to check.

sir-dunxalot:
Is there anything I can do to simplify the test helpers? Is it the logic inside each one that is hard to read? Or maybe the use of more abstracted helpers?

a15n:
> "If we wanted to only have one assertion in assertTooltipSpacing() would you still test for side and spacing or would you test for spacing and ignore side? If we did the latter, then the spacing could be correct (20px from the target) but the side could be completely wrong. Maybe that's how it should function but I just want to check."

The `assertTooltipVisible` [checks for two booleans](https://github.com/sir-dunxalot/ember-tooltips/blob/master/test-support/helpers/ember-tooltips.js#L99) to be true but presents it as one assert. I was thinking something like that.


--------------
> "Is there anything I can do to simplify the test helpers? Is it the logic inside each one that is hard to read? Or maybe the use of more abstracted helpers?"

I think the `expected____Distance` values are a bit hard to understand. Although [this approach](https://github.com/sir-dunxalot/ember-tooltips/pull/145/files#diff-9693234231881212e47f7813b6502a69L128) is more lines I feel like the code is easier to understand. But this isn't a blocker since this is an internal function.

test-support/helpers/ember-tooltips.js
a15n: I think it's possible that people will use this test helper in different browsers (phantom, firefox, etc) so it's possible that the [warning here](https://github.com/sir-dunxalot/ember-tooltips/pull/145/files#diff-b7eb48ff1a009e6b663bff11a464e390L15) will be prophetic.

But I don't think we should make this test looser than necessary. If it's easy let's try this test in different browsers (I think it's a travis thing), if not let's keep it as strict as possible and then loosen it as the community needs it.

sir-dunxalot:
Ok, both of those point make sense. I'll see what I can do to simplify the expected______Distance code too.

sir-dunxalot:
PR review fixes made. This should be ready to merge.

a15n:
LGTM

Issue #138, opened 2016-12-21
lazy-render-wrapper doesn't support target='some-id'


a15n:
The lazy-render-wrapper will only attach event listeners on [the parent](https://github.com/sir-dunxalot/ember-tooltips/blob/dad31a5e5fb04bf85fbc933efbfee3154b5a00d9/addon/components/lazy-render-wrapper.js#L148). This works fine for `<x>-on-element` but breaks when `target='some-id'` and can potentially break when using `<x>-on-component`.

To fix this we need to 
* make sure that the lazy-render-wrapper is attaching event listeners to the correct element
  * needs to work for `<x>-on-element`, `<x>-on-component`, and when `target='some-id'`
* update `some-component` so that it has an intermediate div so it behaves differently than a plain old div

Here is an example of some code that would break...

```
<div>
  <div id='some-id'>
  </div>

  {{tooltip-on-element target='some-id'}}
</div>
```

a15n:
149 fixed this and added specific tests to test the situations described above.

PR #149, opened 2016-12-30
add level to #some-component, proper $target for lazy-render-wrapper, tests

a15n:
this addresses the bugs identified in this ticket https://github.com/sir-dunxalot/ember-tooltips/issues/138

notale improvements...
* lazy-render-wrapper [will now get](https://github.com/sir-dunxalot/ember-tooltips/pull/149/files#diff-8096b1ffac251e79ea18b7393d9f241fR145) the correct $target element, which may be different when using `<x>-on-component`, `<x>-on-element`, and when `target="some-id"`
  * [before](https://github.com/sir-dunxalot/ember-tooltips/pull/149/files#diff-8096b1ffac251e79ea18b7393d9f241fL148) it was always getting the parentElement
* renamed `entryInteractionEvents` to `_lazyRenderEvents`
* added [_targetEventNameSpace](https://github.com/sir-dunxalot/ember-tooltips/pull/149/files#diff-8096b1ffac251e79ea18b7393d9f241fR163) to keep the code DRY
* willDestroyElement also references `$target` to clean up event handlers
* added an empty `<div>` to `some-component`, which adds extra robustness for `-on-component` tests while maintaining the same UI in the dummy site
* added `render-events-test.js`

### DIFF
--- a/addon/components/lazy-render-wrapper.js
+++ b/addon/components/lazy-render-wrapper.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 import layout from 'ember-tooltips/templates/components/lazy-render-wrapper';
 
 const { computed, get, $ } = Ember;
+export const targetEventNameSpace = 'target-lazy-render-wrapper';
 
 // https://github.com/emberjs/rfcs/issues/168
 // https://github.com/emberjs/ember.js/pull/12500
@@ -142,6 +143,14 @@ export default Ember.Component.extend({
 		return _lazyRenderEvents;
 	}),
 
+	/**
+	 * A jQuery element that the _lazyRenderEvents will be
+	 * attached to during didInsertElement and
+	 * removed from during willDestroyElement
+	 * @property $target
+	 * @type jQuery element
+	 * @default the parent jQuery element
+	 */
 	$target: computed('target', 'tetherComponentName', function() {
 		const target = this.get('target'); // #some-id
 		let $target;
@@ -151,7 +160,17 @@ export default Ember.Component.extend({
 		} else if (this.get('tetherComponentName').indexOf('-on-component') >= 0) {
 			// TODO(Andrew) refactor this once we've gotten rid of the -on-component approach
 			// share the functionality with `onComponentTarget`
-			const targetViewElementId = this.get('parentView.elementId');
+			const targetView = this.get('parentView');
+
+			if (!targetView) {
+				console.warn('No targetView found');
+				return null;
+			} else if (!targetView.get('elementId')) {
+				console.warn('No targetView.elementId');
+				return null;
+			}
+
+			const targetViewElementId = targetView.get('elementId');
 			$target = $(`#${targetViewElementId}`);
 		} else {
 			$target = getParent(this);
@@ -160,7 +179,6 @@ export default Ember.Component.extend({
 		return $target;
 	}),
 
-	_targetEventNameSpace: 'target-lazy-render-wrapper',
 	didInsertElement() {
 		this._super(...arguments);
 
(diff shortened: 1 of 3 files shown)


addon/components/lazy-render-wrapper.js
```diff
@@ -120,46 +120,72 @@ export default Ember.Component.extend({
 	_shouldShowOnRender: false,
 
 	event: 'hover', // hover, click, focus, none
-	entryInteractionEvents: computed('event', function() {
+	_lazyRenderEvents: computed('event', function() {
```
sir-dunxalot: I like this improvement!

addon/components/lazy-render-wrapper.js
```diff
@@ -120,46 +120,72 @@ export default Ember.Component.extend({
...
+			_lazyRenderEvents.push('mouseenter');
 		} else if (event === 'click') {
-			entryInteractionEvents.push('click');
+			_lazyRenderEvents.push('click');
+		}
+
+		return _lazyRenderEvents;
+	}),
+
+	$target: computed('target', 'tetherComponentName', function() {
```
sir-dunxalot: Can you add a bit of documentation to describe why this CP is necessary? I like the style of the [YUI Docs Syntax](http://yui.github.io/yuidoc/syntax/index.html) and have tried to use that in a couple of places. What do you think?

sir-dunxalot: And when I say "documentation" I mean code comments documentation and not readme documentation.

addon/components/lazy-render-wrapper.js
```diff
@@ -120,46 +120,72 @@ export default Ember.Component.extend({
...
+		const target = this.get('target'); // #some-id
+		let $target;
+
+		if (target) {
+			$target = $(target);
+		} else if (this.get('tetherComponentName').indexOf('-on-component') >= 0) {
+			// TODO(Andrew) refactor this once we've gotten rid of the -on-component approach
+			// share the functionality with `onComponentTarget`
+			const targetViewElementId = this.get('parentView.elementId');
```
sir-dunxalot: Will there always be a `parentView` here?

addon/components/lazy-render-wrapper.js
```diff
@@ -120,46 +120,72 @@ export default Ember.Component.extend({
...
+			$target = $(`#${targetViewElementId}`);
+		} else {
+			$target = getParent(this);
 		}
 
-		return entryInteractionEvents;
+		return $target;
 	}),
 
+	_targetEventNameSpace: 'target-lazy-render-wrapper',
```
sir-dunxalot: Should this be a `const` in the file instead of a property on the component? That would make it truly private and would still avoid repetitions of the classname in the code.

tests/integration/components/render-events-test.js
```diff
@@ -0,0 +1,127 @@
...
+
+    let eventHandler = eventsObject[eventObjectName][0];
+    assert.equal(eventHandler.origType, event,
+        `the eventHandler's origType property should equal ${event}`);
+
+    assert.ok(eventHandler.namespace.indexOf('target-lazy-render-wrapper') >= 0,
+        'the eventHandler\'s namespace property be unique to ember-tooltips');
+  }
+
+  function assertNumEventsExist(num) {
```
sir-dunxalot: I think the use of these helpers is self-explanatory but would like at least one line of comments to explain why and when these function should be used. This would enabled contributors to more easily understand and add to these files.

addon/components/lazy-render-wrapper.js
a15n: picking variable names is hard. I'm happy with this one though

addon/components/lazy-render-wrapper.js
a15n: I think that's a good idea. I'll add some docs here in the YUI style. I think we should go through all the code and add docs to the complex properties and methods at some point.

addon/components/lazy-render-wrapper.js
a15n: I just copied over the console.warn checks from the utils file. Better safe than sorry. I'm OK with copying because I'll be cleaning this up when we get rid of the `<x>-on-component` components in favor of `ember-tooltip` with a boolean.

a15n:
Thanks for the review. I've addressed the feedback here. I think this is ready to merge.

https://github.com/sir-dunxalot/ember-tooltips/pull/149/commits/56eedd6da6c3712077dc2d0f0ef12132c2acb47e

sir-dunxalot:
Released as `2.6.1`.

sir-dunxalot:
[Release notes here](https://github.com/sir-dunxalot/ember-tooltips/releases/tag/2.6.1).

Issue #150, opened 2016-12-30
race condition when enableLazyRendering and target="some-input"


a15n:
The tooltip will appear then immediately disappear when the input is clicked on...

```
<input id="some-input">

{{tooltip-on-element
    target="#some-input"
    text="hello world"
    event="click"
    enableLazyRendering=true
}}
```

I believe this has something to do with focus events occurring before click events... will investigate

![race condition](https://cloud.githubusercontent.com/assets/7050871/21558171/75a17f0a-cdea-11e6-89f2-31958d5256bc.gif)

PS: I saw the same thing for popover-on-element

a15n:
@jiongbozhang  FYI

a15n:
it seems like we can get around this if we use `event='focus'` (as is shown in the dummy app). Should we try to make a race condition fix for this or just encourage people to use focus?

sir-dunxalot:
We should fix the race condition because this is a common use case for tooltips in Ember.

sir-dunxalot:
#149 is now on master.

a15n:
PR here https://github.com/sir-dunxalot/ember-tooltips/pull/153

sir-dunxalot:
Fixed by `2.7.3`.

PR #153, opened 2016-12-31
A15n/popover focus fix

a15n:
This PR fixes these three issues related to inputs and adds tests...

1. When a tooltip was attached to an input we used to see a flicker in this instance
```
<input id="some-input">
{{tooltip-on-element 
  event="click" 
  target="#some-input" 
  enableLazyRendering=true
}}
```
2. When a popover was attached to an input we used to see a flicker in this instance
```
<input id="some-input">
{{popover-on-element
  target="#some-input"
  event="click"
  enableLazyRendering=true
}}
```
3. When a popover was attached to an input we used to not be able to click away to close it in this instance
```
<input id="some-input">
{{popover-on-element 
  event="focus" 
  target="#some-input"
  enableLazyRendering=true
}}
```

Initial ticket -> https://github.com/sir-dunxalot/ember-tooltips/issues/150


a15n:
@sir-dunxalot this is ready for review thanks

addon/components/lazy-render-wrapper.js
```diff
@@ -206,6 +206,7 @@ export default Ember.Component.extend({
 				} else {
 					this.set('_hasUserInteracted', true);
 					this.set('_shouldShowOnRender', true);
+					this.set('_isInProcessOfShowing', true);
```
sir-dunxalot: So I've been thinking about our using of private variables on components given the other less-used alternative that is truly private variables in the file. For example:

```js
import Ember from 'ember';

let isInProcessOfShowing = false;

export default Ember.Component.extend({

  ...

   someMethod() {
      ...

      isInProcessOfShowing = true;
   }

});
```

What is your preferred approach? The example above would probably not work is we need to use `_isInProcessOfShowing` in other modules (e.g. a component that extends `lazy-render-wrapper`) so it may not be feasible. However, I would like to at least answer the question of which method we prefer.

addon/components/tether-tooltip-on-element.js
```diff
@@ -25,7 +25,14 @@ export default TooltipAndPopoverComponent.extend({
 
       if (_showOn === _hideOn) {
         $target.on(_showOn, () => {
-          this.toggle();
+          // When using enableLazyRendering the focus event occurs before the click event.
```
sir-dunxalot: Nice commentary 👍

addon/templates/components/lazy-render-wrapper.hbs
```diff
@@ -3,8 +3,10 @@
 		passedPropertiesObject=passedPropertiesObject
 		class=class
 		id=id
+		classNames=classNames
+
 		_shouldShowOnRender=_shouldShowOnRender
-    classNames=classNames
+		_isInProcessOfShowing=_isInProcessOfShowing
```
sir-dunxalot: If we're passing private properties in to a child component, should they be private? I'd argue that `isInProcessOfShowing` should be a public property *or* a private property on the `{{component}}`.

sir-dunxalot:
Reviewed. I left a couple of lingering questions. #154, which is about to be merged, will cause a couple of failures with code conventions so I expect you'll need to make an addition in order to have a passing test suite. If we don't like the conventions that #154 added we can, of course, change them to suit.

addon/components/lazy-render-wrapper.js
a15n: I think that the use of private variables is the best approach when used within a single file. But when passed from component to component we have to pass them, as you've identified.

addon/templates/components/lazy-render-wrapper.hbs
a15n: My thoughts are that properties should be private if the user never needs to modify them to use the addon. In this case a user should never modify _isInProcessOfShowing so I think it should be private.

a15n:
@sir-dunxalot thanks for your review. This has been merged with master and is ready to merge.

Issue #151, opened 2016-12-30
Improve code conventions and linting


sir-dunxalot:
Our addon follows some code conventions loosely. Parts of the code are messy and inconsistent with one another. We should tidy up the code (conventions TBD) whilst using something like [ember-cli-eslint](https://github.com/ember-cli/ember-cli-eslint) to make it easy for contributors to work with our chosen code conventions.

sir-dunxalot:
@a15n What do you think? I have avoided adding comments for code conventions in PRs but I'd like to improve consistency. I don't think this is a big issue and am fine if we don't want to focus on styling right now (we can close this issue without making any changes because this is a rather opinionated improvement and not a functionality addition).

a15n:
I think eslint is a great idea. Can we do this before 3.0.0?

sir-dunxalot:
Great. I have added linting to the pre-3.0 checklist in #109.

sir-dunxalot:
I have begun work on this on the [code-linting](https://github.com/sir-dunxalot/ember-tooltips/tree/code-linting) branch.

sir-dunxalot:
Fixed by #154

PR #154, opened 2017-01-01
Code linting

sir-dunxalot:
This PR adds code linting to this addon using the following third-party code:

- [ember-cli-eslint](https://github.com/ember-cli/ember-cli-eslint) for the linting, which adds a test to the QUnit test suite for every JS file in the addon
- [eslint-plugin-ember-suave](https://github.com/DockYard/eslint-plugin-ember-suave), which adds the default eslint configuration and some custom rules designed for Ember
- [eslint-plugin-netguru-ember](https://github.com/netguru/eslint-plugin-netguru-ember), which adds some more custom rules designed for Ember

Where our preferred code conventions differs from the eslint plugins above, overriding configuration is  defined in the [`.eslintrc.js` file here](https://github.com/sir-dunxalot/ember-tooltips/blob/code-linting/.eslintrc.js).

The only rules I wanted to add that isn't yet supported are 1) enforcing empty lines before and after blocks and 2) enforcing empty lines before and after computed properties and methods. As such, I expect some PR reviews will still touch on issues on code conventions and styling.

You can see the eslint tests being run on Travis [in this job](https://travis-ci.org/sir-dunxalot/ember-tooltips/jobs/188082028).


sir-dunxalot:
Fixes #151

.jscsrc
```diff
@@ -1,80 +0,0 @@
-{
```
sir-dunxalot: Removed because we don't use jshint anymore.

.jshintrc
```diff
@@ -1,32 +0,0 @@
-{
```
sir-dunxalot: Removed because we don't use jshint anymore.

addon/components/tether-popover-on-element.js
```diff
@@ -4,53 +4,96 @@ import layout from 'ember-tooltips/templates/components/tether-popover';
```
sir-dunxalot: This file ([full version here](https://github.com/sir-dunxalot/ember-tooltips/blob/93505b486c93a3cde1095fc2084aee3fdee7b21d/addon/components/tether-popover-on-element.js)) is a good example of what code conventions we're actually enforcing. The conventions and rules are too numerous to list.

sir-dunxalot:
Thank you. I thought about using the `eslint --fix` option but due to installation complications where I wanted to avoid installing `eslint` globally I opted to make all the changes manually.

Issue #158, opened 2017-01-06
unable to use popover in IE11 (windows 8.1)


ktelles:
we're having a problem where the popover appears correctly but when you click on it it interacts with the content underneath it, instead of the popover itself. this is on IE11 (windows 8.1). thanks so much!

![bug](https://cloud.githubusercontent.com/assets/22331782/21736422/6fff7e76-d425-11e6-9c2c-9a7f68d50da5.gif)

a15n:
https://github.com/sir-dunxalot/ember-tooltips/pull/159

sir-dunxalot:
Fixed by `2.7.3`.

PR #159, opened 2017-01-07
pointer-events: auto; is the correct CSS to use

a15n:
https://github.com/sir-dunxalot/ember-tooltips/issues/158

`pointer-events: initial;` was not working in IE. Setting it to `initial` was a mistake and I always wanted the behavior from the `auto` value.

https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events

### gif in IE after fix
popover in ie

### DIFF
--- a/addon/styles/addon.scss
+++ b/addon/styles/addon.scss
@@ -18,7 +18,7 @@
 }
 
 .ember-popover[aria-hidden="false"] {
-  pointer-events: initial;
+  pointer-events: auto;
   cursor: initial;
   -webkit-touch-callout: auto;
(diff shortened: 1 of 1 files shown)


a15n:
PS: I also checked that this still works in Chrome.

a15n:
fyi @ktelles

sir-dunxalot:
LGMT. Will release as a patch today.

sir-dunxalot:
Released as `2.7.3`.

sir-dunxalot:
Released as `2.7.3`.

PR #162, opened 2017-01-08
Absolute parent fix

sir-dunxalot:
This PR makes tooltips render as a child of the Ember app's `rootElement` if it's defined. When `rootElement` is not defined, tooltips will still render as a child of the body.

This change fixes event delegation issues in test suites where events and actions were fired but not caught by the Ember app of parent components. This fixes #157 and #141.

PR #169, opened 2017-01-20
Bump ember-tether version to maintain compatibility with other libraries

mfeckie:
Some other libraries have bumped their requirements e.g. `ember-shepherd` which causing npm failures when using both.

This brings `ember-tether` to 0.4.0 and the tests seem to pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "ember-export-application-global": "^1.0.5",
     "ember-load-initializers": "^0.5.1",
     "ember-resolver": "^2.0.3",
-    "ember-tether": "0.4.0",
+    "ember-tether": "0.3.1",
     "eslint-plugin-ember-suave": "^1.0.0",
     "eslint-plugin-netguru-ember": "^1.6.5",
     "loader.js": "^4.0.1"
@@ -61,7 +61,7 @@
     "ember-hash-helper-polyfill": "^0.1.1"
   },
   "peerDependencies": {
-    "ember-tether": "0.4.0"
+    "ember-tether": ">=0.3.1 <=0.4.0"
   },
   "ember-addon": {
     "configPath": "tests/dummy/config",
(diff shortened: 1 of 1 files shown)


a15n:
What if we change the [peerDependency](https://github.com/sir-dunxalot/ember-tooltips/blob/master/package.json#L64) to `>=0.3.1 <=0.4.0` (0 prefixed versions don't support `^`)?  This would support both versions.

Here is the 0.4.0 release. https://github.com/yapplabs/ember-tether/releases/tag/v0.4.0

We can't drop support for 0.3.1 because newer versions of ember-tether don't explicitly support ember 1.13

a15n:
Just tried this locally. When I kept the installed version of ember-tether as 0.3.1 but updated the peerDependencies to include 0.4.0 all of the builds passed. @mfeckie can you try upgrading your peerDependency requirement as I recommended and see if that works for you?

What version of Ember are you on?

mfeckie:
@a15n Done.

nagits:
> We can't drop support for 0.3.1 because newer versions of ember-tether don't explicitly support ember 1.13, which we do support

One possible solution:
1. Remove ember-tether from peerDependencies.
2. Add default blueprint:
```js
// ember-tooltips/blueprints/ember-tooltips/index.js
var VersionChecker = require('ember-cli-version-checker');

module.exports = {
  ...
  afterInstall: function() {
    var checker = new VersionChecker(this);
    var ember = checker.forEmber();
    return addon.addAddonsToProject({ 
      packages: [
        { name: 'ember-tether', target: ember.isAbove('2.0.0') ? '>=0.3.1 <=0.4.0' : '0.3.1' }
      ]
    });
  }
}
```

sir-dunxalot:
Please see the discussion in #162, which is also hoping to update this library to ember-tether `0.4.0`.

sir-dunxalot:
I'm closing this in favor of #162, which uses the [new ember-tether release](https://github.com/yapplabs/ember-tether/pull/41) and ensures the test suite supports all Ember versions we want, including Ember 1.13.

Thank you for opening the PR though.

Issue #122, opened 2016-12-07
Glimmer 2 DOMException for popover components


jordpo:
With Ember 2.10, an exception is thrown if you are wrapping a popover component in an if block. It seems related to this issue:

https://github.com/yapplabs/ember-wormhole/issues/66

We are getting around it by wrapping the popover- component in a div.

a15n:
LGTM

@sir-dunxalot do you approve?

sir-dunxalot:
Was that meant to be a comment on #176? If so, yes.

a15n:
haha yes it was

maxfierke:
Addressed in 2.11.1.

PR #176, opened 2017-01-31
Fix glimmer 2 exceptions

Glennvd:
Fix for https://github.com/sir-dunxalot/ember-tooltips/issues/122 that also works for tooltip-on-element

### DIFF
--- a/addon/components/tether-tooltip-and-popover.js
+++ b/addon/components/tether-tooltip-and-popover.js
@@ -216,7 +216,7 @@ export default EmberTetherComponent.extend({
       return null;
     }
 
-    const parentElement = $element.parent();
+    const parentElement = $element.parent().parent();
     let parentElementId = parentElement && parentElement.attr('id');
 
     if (!parentElementId) {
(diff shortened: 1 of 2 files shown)


sir-dunxalot:
@a15n What do you think?

a15n:
LGTM after the tests pass (I just triggered them)

a15n:
@Glennvd it looks like tests are failing

sir-dunxalot:
How can we add test coverage for this?

Glennvd:
I believe the tests need to be updated to properly detect if the tooltips again, I'll try to look into it if I find some time

addon/components/tether-tooltip-and-popover.js
```diff
@@ -216,7 +216,7 @@ export default EmberTetherComponent.extend({
       return null;
     }
 
-    const parentElement = $element.parent();
+    const parentElement = $element.parent().parent();
```
a15n: I think this will break the use when a `tether-tooltip-and-popover` is used by itself. 

```
<div>
  hover here
  {{tether-tooltip-and-popover text="foo"}}
</div>
```

maxfierke:
Closing in favor of #265

PR #178, opened 2017-02-01
Add postcss to dummy app

sir-dunxalot:
**Requires #177.**

This PR adds postcss and autoprefixer to make the dummy app's CSS more cross-browser compatible after #177 removed compass-boilerplate as a dependency.

PR #179, opened 2017-02-02
Postcss added

sir-dunxalot:
It seems there was an issue with #178 and it didn't update the correct files. This is an addition to make that PR meaningful.


sir-dunxalot:
Merging this through quickly to resolve the lack of postcss currently on master.

PR #183, opened 2017-02-06
the effect name is in fact `slide`, not `hover`

humanchimp:
This PR just fixes what seems to be a typo in a test


a15n:
Good catch thanks!

PR #186, opened 2017-03-06
Fixes issue where toggle is called after show which hides the tooltip…

MiguelMadero:
…/popver.

Sets _isInProcessOfShowing from `didInsertElement` before showing.

### DIFF
--- a/addon/components/tether-tooltip-and-popover.js
+++ b/addon/components/tether-tooltip-and-popover.js
@@ -317,6 +317,12 @@ export default EmberTetherComponent.extend({
     this._super(...arguments);
 
     if (this.get('_shouldShowOnRender')) {
+
+      /* When using enableLazyRendering the focus event occurs before the click event.
+        When this happens we don't want to call focus then click.
+        _isInProcessOfShowing prevents that from happening. */
+
+      this.set('_isInProcessOfShowing', true);
       this.show();
     }
 


a15n:
@MiguelMadero could you also add a test to protect this use case?

MiguelMadero:
Fair enough! I was sort of expecting you would ask :D  I started writing the test, but it was a bit complicated to create a repro state outside of our app. I'll have another stab at it and update the PR.

sir-dunxalot:
@MiguelMadero master branch will receive breaking changes within a week or two, making this PR unmergeable after that. This is because the ember-tooltip's `3.0.0` will soon be on master. The `3.0.0` release is a major refactor, which fixes this issue so there is no `_isInProcessOfShowing` property in `3.x`.

So we have two choices here:

- Merge this PR to master within a couple of days (requiring a test for the change) so it's in the `2.x` codebase before `3.x` is merged to master
- Close this PR and, if desired, make this change on a soon-to-exist ember-tooltips `2.x` branch, which will accept improvements to `2.x` of this library

I will come back to this issue in a few days and will close it if you opt for the latter or don't have the time to update the PR. Either way is fine by me. Thank you for the contribution regardless!

MiguelMadero:
@sir-dunxalot apologies for never closing this. At this point, it's probably better to just close it and focus on 3.0. 

Thanks for the detailed information!

sir-dunxalot:
No problem, sounds good to me.

Issue #72, opened 2016-05-31
Visual issues with tooltips


hoboman313:
First off, this addon looks very promising and is much easier to use with ember than something like bootstrap tooltips. Secondly...I ran into a bunch of visual issues that I was hoping I could get some help with. I hope it's okay that I list several issues here.

1.)  keepInWindow doesn't actually try to keep the tooltip inside the window...or maybe it just doesn't try as hard as the bootstrap tooltip addon?
An example: When I resize the window, the tooltip doesn't get resized to try to fit in. <a href="http://imgur.com/KRco44M"><img src="http://i.imgur.com/KRco44M.png" title="source: imgur.com" /></a>
If you look at the bootstrap example ( http://getbootstrap.com/javascript/#tooltips ) while resizing the page, you'll notice that the tooltip overflows to multiple lines and it does a pretty good job, to a certain extent, to stay inside the window.

2.) ember-tooltips don't get re-rendered between model changes
I've noticed that ember-tooltips always display well after I transition between different routes. However, in the case that I stay on the same route and substitute in different models, the tooltips stop showing up all together.
^ I hope this was descriptive enough. I think what I'm looking for is a functionality to force the ember-tooltips to re-init themselves.

3.) ember-tooltips don't always get displayed under the correct element even when setting a target
Consider the block like
`{{#if yelpData.haveReviews}}
                    <a id="yelp-anchor" href="{{yelpData.url}}" target="_blank" rel="noopener noreferrer">
                      <img src="{{yelpData.rating_img_url_large}}" id="yelp-rating" />
                    </a>
                    {{#tooltip-on-element class="ember-tooltip" target="#yelp-anchor"}}{{yelpData.rating}} based on {{yelpData.review_count}} reviews{{/tooltip-on-element}}
                  {{/if}}`

where the tooltip is attached to #yelp-anchor. When I hover over the element the tooltip correctly comes up, but its position is way off.
<a href="http://imgur.com/jZHmxlY"><img src="http://i.imgur.com/jZHmxlY.png" title="source: imgur.com" /></a>
<a href="http://imgur.com/P4wPeWo"><img src="http://i.imgur.com/P4wPeWo.png" title="source: imgur.com" /></a>
Any help debugging this would be appreciated

sir-dunxalot:
Hi Egor, thanks for the details!

[...]

PR #189, opened 2017-04-07
make sure tooltips stay in window

kybishop:
Partially addresses #72

Prevents tooltips from being cut off if they are near a border of the viewport. Grep for "pin" [here](http://tether.io/) for explanation of the pin option.

### DIFF
--- a/addon/components/tether-tooltip-and-popover.js
+++ b/addon/components/tether-tooltip-and-popover.js
@@ -181,6 +181,7 @@ export default EmberTetherComponent.extend({
         {
           to: 'window',
           attachment: 'together',
+          pin: true,
         },
       ];
     }


panthony:
Any update about this PR? I'm currently hitting this issue where tooltip and popover are rendered partially outside the viewport.

kybishop:
@panthony not sure what the holdup is here. I ended up making an alternative tooltip library due to this issue in particular.

panthony:
@kybishop What did you use instead? I only need the "popover" kind of element, not the "tooltip" one.

For now I extended the `tether-popover-on-element` component to override the `constraints` property on my app to avoid forking the project just for this.

kybishop:
@panthony https://github.com/kybishop/ember-attacher

nathancarnes:
@kybishop thank you! This fixed worked great for us.

@sir-dunxalot I would love to help get this change merged. Are you open to a rebased PR for it? Would you prefer it's a new bit of configuration?

sir-dunxalot:
I've experimented with this solution locally after seeing @kybishop's insight. I agree that it's a partial solution - there are some window positioning issues it doesn't solve and I haven't yet figured out why. Generally, ember-tether seems to have caused more issues than it fixed for v2. A rewrite of the code that attaches tooltip to element is required here, IMO, to fix this and similar positioning issues. 

I would be happy to push through what fixes we can for this library in lieu of the time required to rewrite the positioning code this library uses ember-tether to achieve.

@nathancarnes, yes, definitely open to a rebased PR for this particular fix.

andreafrost:
@sir-dunxalot I just submitted a pull request that allows users to set an _optional_ parameter to set the `pin`.  That way it won't actually change anything for people who don't want to set `pin: true`:

https://github.com/sir-dunxalot/ember-tooltips/pull/220

(shout out to @kybishop for coming up with a solution that works!)

nathancarnes:
@sir-dunxalot awesome, it looks like @andreafrost beat me to it 😃 

@kybishop do you think it makes sense to close this PR in favor of #220, which implements this same change as an optional parameter?

maxfierke:
Closing in favor of #220, which has been merged in. Thanks!

PR #198, opened 2017-05-24
Add `isDisabled` property to tooltips, allowing them to be disabled

cbargren:
Usage:

```
<button>
  Click me
  {{#tooltip-on-element isDisabled=true}}
    I'm a popup but you can't see me
  {{/tooltip-on-element}}
</button>
```

If there's interest in merging this, I can add tests as needed to get it up to par.


mehanika:
This would be very useful 👍

PR #199, opened 2017-05-29
Wrap jquery event handlers in Ember.run

joukevandermaas:
Without this, sometimes our application's tests will fail with this error:

> Assertion Failed: You have turned on testing mode, which disabled the run-loop's autorun. You will need to wrap any code with asynchronous side-effects in a run.


maxfierke:
Closing in favor of #267, which has been merged.

PR #200, opened 2017-05-30
Fixed a bug in lazy-render-wrapper that causes tooltip context not sent to action handlers

jinkea:
The context parameter was lost when calling the action handlers from lazy-render-wrapper.

also wrote some integration tests for the fix.


jinkea:
@sir-dunxalot can you help take a look and see if you can have this merged soon?

maxfierke:
I believe this should be fixed as of 2.11.0. Feel free to re-open and rebase if that's not the case. Thanks!

PR #206, opened 2017-06-16
Fixes #205 by moving scss to a dev dependency (to keep for dummy app)

### LINKED ISSUE #205: Remove dependency on sass
Hey! First off, thanks for this awesome addon. I would like to request that you remove all dependencies on sass. Our project doesn't use sass, and this is the only addon we use that does. Unfortunately, through no fault of your own, many of us get errors with node-sass and yarn. 

This addon pulls in ember-cli-sass which pulls in broccoli-sass which pulls in node-sass. It's got *very* little sass in this addon. What do you think about just converting the small amount of sass to CSS, and removing the dependency? This would speed up dependency installation, remove a build step, and save *lots* of downloads of all three of those libraries for projects that don't really need them.

xdumaine:
Convert addon.scss to css, move scss to dev dep


noxan:
Any change of getting this merged soon? It is the last ember plugin relying on `sass` we have in our codebase and `sass` breaks our complete setup due to its native bindings.

PR #210, opened 2017-07-12
add support for ember-engines

andrewfan:
When using inside ember-engines baseConfig does not have APP property


maxfierke:
closing in favor of #257. Thanks!

PR #228, opened 2017-12-11
Fix events listeners removal on Destroy

Vincz:
In the `_tooltipEvents` property, the callback is registered (to be able to remove listeners later). But the callback is wrapped when added to the target. So, the removal won't work onDestroy.
We should either remove the wrapper or register it in properly in the `_tooltipEvents` list.


snewcomer:
@sir-dunxalot Seems like lots of PRs out there for this lib.  Would you like nay help with this?

PR #229, opened 2017-12-14
Prevents Fastboot error in willDestroy hook

willviles:
This PR fixes the Fastboot error detailed below inside the `willDestroy` hook.

```
There was an error running your app in fastboot. More info about the error:
 TypeError: Cannot read property 'dispose' of null
    at Class.willDestroy (/Users/will/Code/themarshalsgroup-riskmap/tmp/broccoli_persistent_filterautoprefixer_filter-output_path-6YxFbSBF.tmp/assets/addon-tree-output/ember-tooltips/components/ember-tooltip-base.js:205:1)
```

It simply checks for the presence of a tooltip before running any teardown functions.

### DIFF
--- a/addon/components/ember-tooltip-base.js
+++ b/addon/components/ember-tooltip-base.js
@@ -188,6 +188,12 @@ export default Component.extend({
 
   willDestroy() {
     this._super(...arguments);
+    const tooltip = this.get('_tooltip');
+
+    if (!tooltip) {
+      return;
+    }
+
     this.hide();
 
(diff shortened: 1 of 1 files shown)


mhluska:
I'm having the same problem. Is it there any progress on this?

maxfierke:
3.0.0-beta.3 moved away from `willDestroy` and is using `willDestoryElement` instead, which should not be called under Fastboot. It might be worth trying out 3.0.0-beta.3 and seeing if this is still an issue (there may be other issues related to #271 that appear, however)

PR #235, opened 2018-01-15
[3.0 beta] Call `setSpacing()` on popper.js create instead of update

lan0:
`popperInstance.state.updateBound();` triggers `onUpdate()` which leads to an endless loop and 100% cpu utilization.

### DIFF
--- a/addon/components/ember-tooltip-base.js
+++ b/addon/components/ember-tooltip-base.js
@@ -333,6 +333,7 @@ export default Component.extend({
                   /* Once the wormhole has done it's work, we need the tooltip to be positioned again */
 
                   run.scheduleOnce('afterRender', () => {
+                    this.setSpacing();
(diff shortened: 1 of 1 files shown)


maxfierke:
Thanks for looking into this! One thing I ran into testing this, was that this changeset prevents the spacing from being updated on open tooltips and popovers, which can be a problem if the tooltip/popover is open while scrolling and needs to change position (e.g. flipping sides due to nearing the edge of the viewport)

I've opened #268 as an alternative approach to solving this that preserves the updating behavior and fixes it for the cases I was reproducing for, while avoiding the thrashing issue. If you're still encountering issues, I'd be interested in your take on it, as to whether it fixes the issue for your use cases, as well.

lan0:
This fixes the issue for me, thanks! Looking forward to the release of 3.0.

PR #240, opened 2018-02-21
Fix issue with code running outside the run loop

alexlafroscia:
I was seeing some flakey tests around that cleanup work that was scheduled on `focusout` since there wasn't necessarily a runloop present when the work was scheduled. This change fixed my issue.

### DIFF
--- a/addon/components/tether-tooltip-and-popover.js
+++ b/addon/components/tether-tooltip-and-popover.js
@@ -32,6 +32,11 @@ function cleanNumber(stringOrNumber) {
   return cleanNumber;
 }
 
+function stopTether() {
+  this.set('_isTetherEnabled', false);
+  this.get('_tether').disable();
+}
+
 export default EmberTetherComponent.extend({
 
   passedPropertiesObject: null,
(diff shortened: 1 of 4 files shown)


alexlafroscia:
cc @jasonmit

addon/components/tether-tooltip-on-element.js
```diff
@@ -62,24 +62,20 @@ export default TooltipAndPopoverComponent.extend({
...
         }
 
-        $target.focusout(() => {
-          this.hide();
-        });
+        $target.focusout(run.bind(this, this.hide));
       }
 
-      $target.keydown((keyEvent) => {
+      $target.keydown(run.bind(this, (keyEvent) => {
```
alexlafroscia: I didn't specifically have an issue here but all of these event handlers should really ensure their work is being done in a run loop.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "ember-tooltips",
```
maxfierke: could you omit these `package.json` and `package-lock.json` changes?

addon/components/tether-tooltip-and-popover.js
```diff
@@ -530,18 +537,7 @@ export default EmberTetherComponent.extend({
   },
 
   stopTether() {
-    run.schedule('afterRender', () => {
-      if (!this.isDestroyed && !this.isDestroying) {
```
maxfierke: I think we probably need to keep these checks in place.

package.json
alexlafroscia: Yup, for sure.

We added some changes outside the scope of this PR so that I could publish my updates to NPM since installing from Github is not stable.

I can close this PR and make a new one with just the changes we actually want to make without my additional changes.

alexlafroscia:
Closed in favor of #267

PR #245, opened 2018-04-18
Positioning Issue with 3.0.0-beta.2

mfees:
I was having an issue with 3.0.0-beta.2 where if a higher container element had `overflow-y: scroll` and the parent element to the tooltip was `display: flex`, the tooltip would render over the parent element regardless of the positioning I passed into the component.

I recreated an example of this [here](https://mfees.github.io/ember-tooltip-example/example/).

In the modifiers in ember-tooltip-base.js, I added this:
```javascript
preventOverflow: {
   escapeWithReference: true
}
```
This seems to resolve the issue and didn't seem to cause additional issues when overflow and flex are not set on the parents.

### DIFF
--- a/addon/components/ember-tooltip-base.js
+++ b/addon/components/ember-tooltip-base.js
@@ -317,6 +317,9 @@ export default Component.extend({
                 flip: {
                   enabled: this.get('enableFlip'),
                 },
+                preventOverflow: {
+                  escapeWithReference: true
+                }
               },
 
               onCreate: (tooltipData) => {


sir-dunxalot:
I have not used/tried the `escapeWithReference` property whilst writing the `3.x` branch but am fine with merging this change since:

- Manual tests by @mfees and @maxfierke show no adverse visual effects of common use cases
- The [Popper.js documentation](https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/popper-documentation.md#modifierspreventoverflow) corroborates the desired functionality
- If we get bug reports because of this change, it will be relatively quick and easy for us to remove this as a default popper option and, instead, add a property you can set on the components.

This might look **something** like this:

```
{{ember-tooltip preventOverflow=true}}Additional info here{{/ember-tooltip}}
```

PR #246, opened 2018-04-26
Fix for missing addon.sass file

### LINKED ISSUE #243: File not found: \addon.sass Build Error
https://github.com/aexmachina/ember-cli-sass#alternative-addon-usage

Lack of obeying this advice is causing my addon to fail after installing ember-tooltips.
Can I ask you to move styles to appropriate place?

lamabiker:
- Fixes #243 
- Branched off 2.9.2
- Added an `addon.sass` file that's a duplicate of `addon.scss`
- Bumped `ember-cli-sass` to v7.2.0


maxfierke:
@lamabiker Hey, sorry about this, but I think we'll likely close this PR in favor of #264, which will move away from SASS/SCSS to just plain CSS. Thanks for your work on this! The good news is that it's my understanding that due to the way imports work in ember-cli-sass/node-sass, the plain CSS should _Just Work™_ and this issue should be fixed once that PR is merged. (I will verify before we release.)

lamabiker:
No problem! If there's no plan to let the user import styles directly into their app, moving to plain CSS definitely makes sense.

Issue #247, opened 2018-05-03
v2.9.2 broken on ember-beta


jamesarosen:
Whenever I use a `tooltip-on-component`, I get

> Error: Assertion Failed: Attempted to call .sendAction() with the action 'onDestroy' on the destroyed object '<dummy@component:tether-tooltip-on-component::ember3418>'.

from [TooltipOnComponent#willDestroy](https://github.com/sir-dunxalot/ember-tooltips/blob/2.9.2/addon/components/tether-tooltip-and-popover.js#L516-L518)

mmadsen2:
I'm not sure if this is the best fix for this, but I was able to resolve this problem by changing `onDestroy` to `onDestroyElement` in tether-tooltip-and-popover.js.

maxfierke:
This is now fixed in master. Will be in 2.10.0 release. Thanks for the report!

PR #258, opened 2018-07-18
Fix: Error in Ember 3.0 when onDestroy action is sent

mattgraham1995:
Addresses #247 

In Ember 3.0, the `sendAction` method now has an assert in it that checks for `this.isDestroyed` and `this.isDestroying`. The `willDestroy` hook is reached after `this.isDestroying` is set to true. The `hide()` method would just return if `isDestroying` was true, but since we are calling it when it is no longer true, we just check if the element is shown instead.

### DIFF
--- a/addon/components/tether-tooltip-and-popover.js
+++ b/addon/components/tether-tooltip-and-popover.js
@@ -499,14 +499,16 @@ export default EmberTetherComponent.extend({
     }
   },
 
-  willDestroy() {
+  willDestroyElement() {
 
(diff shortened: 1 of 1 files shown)


mattgraham1995:
@sir-dunxalot Can you please give this a look soon? My team is wanting to upgrade to Ember 3.0, but this error is causing our builds to fail.

DanielOchoa:
Glad to at least find a quick fix branch, even if the owner has gone awol. Thanks @mattgraham1995

Issue #113, opened 2016-11-19
Bug: 404 not found for ember-tooltips.css.map


mayatron:
The source map request to `/assets/ember-tooltips.css.map` is 404 not found for a clean build with the latest ember cli 2.9.1. Similar to an earlier closed issue https://github.com/sir-dunxalot/ember-tooltips/issues/76.

Steps to reproduce:

```
ember new myapp && cd myapp
ember install ember-tooltips && npm uninstall ember-welcome-page --save-dev
ember s
```

![screen shot 2016-11-19 at 2 33 53 pm](https://cloud.githubusercontent.com/assets/1009725/20458872/bb5596f2-ae65-11e6-87ed-def193475c4f.png)

![screen shot 2016-11-19 at 2 40 11 pm](https://cloud.githubusercontent.com/assets/1009725/20458913/5e835b48-ae66-11e6-9b5f-37940d258e8c.png)

a15n:
I have `ember-cli:  2.5.0` but as soon as I created an app I updated to 2.9.1 (not sure that changes anything).

I tried reproducing this but was unable to. The page loads fine for me.

sir-dunxalot:
I believe this is a duplicate of #76, which was identified as an upstream issue.

Issue #76, opened 2016-06-08
Ember-tooltips.css.map is missing


Deovandski:
Has anyone else came across this issue where something is requesting for **/assets/ember-tooltips.css.map**? I am using Ember-cli-rails by the way.

```
Started GET "/assets/ember-tooltips.css.map" for 127.0.0.1 at 2016-06-08 02:14:41 -0500
02:14:41 web.1  | 
02:14:41 web.1  | ActionController::RoutingError (No route matches [GET] "/assets/ember-tooltips.css.map"):
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/action_dispatch/middleware/debug_exceptions.rb:21:in `call'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/action_dispatch/middleware/show_exceptions.rb:30:in `call'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/rails/rack/logger.rb:38:in `call_app'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/rails/rack/logger.rb:20:in `block in call'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/active_support/tagged_logging.rb:68:in `block in tagged'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/active_support/tagged_logging.rb:26:in `tagged'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/active_support/tagged_logging.rb:68:in `tagged'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/rails/rack/logger.rb:20:in `call'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/action_dispatch/middleware/request_id.rb:21:in `call'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/rack/methodoverride.rb:22:in `call'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/rack/runtime.rb:18:in `call'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/active_support/cache/strategy/local_cache_middleware.rb:28:in `call'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/rack/lock.rb:17:in `call'
02:14:41 web.1  |   /usr/lib/ruby/vendor_ruby/action_dispatch/middleware/static.rb:120:in `call'
[...]

PR #264, opened 2018-08-29
Refactored css to avoid need for sass, postcss and autoprefixer

chrism:
As per my slack question this is a PR which hopefully addresses issues relating to SASS compilation.

For me, the main one was the missing source map error message which was caused upstream by `ember-cli-sass` https://github.com/sir-dunxalot/ember-tooltips/issues/76 & https://github.com/sir-dunxalot/ember-tooltips/issues/113

But I think it could also simplify a lot of the issues people have had in general working with the styling.

Initially I just flattened the selectors in the `addon.scss` file to remove the need to use any processing and converted it to a css file.

My plan was to then just use processing for the styling of the dummy app.

Unfortunately this wasn't sufficient to use `ember-cli-sass` purely as a dev dependency (because the `addon.scss` file gets automatically included in the pipeline by default when `ember-cli-sass` is installed).

Looking at the styling for the dummy app it showed that SASS was only necessary for some nesting of selectors (and some color variables).

I used the same approach as with the addon.scss file to flatten the selectors and by using native CSS variables it removed the need for SASS completely.

My rationale is that as this styling isn't even used by the addon, purely for the tests and documentation it would be acceptable to require modern browser support.

Doing this allowed me to remove 3 styling addons which are no longer necessary

- `ember-cli-sass`
- `autoprefixer`
- `ember-cli-postcss`

Additionally, in the course of updating the files I noticed that the `ember-code-snippet` addon was injecting some CSS into the `addon.css` that wasn't being used by the addon and actually conflicted with the extra CSS from `_code.scss` in the dummy app.

The current live documentation http://sir-dunxalot.github.io/ember-tooltips/ seems to have some inconsistent code styling and I feel like this might have been a contributing factor.

So I took the opportunity to also clean this up a bit by removing all that unused styling code and then implemented a simple colour scheme that works with the handlebars code (see example below).

screen shot 2018-08-29 at 16 06 43

I also followed the recommendations in the documentation to not use the default `highlight.js` file and instead a custom one.
 
The tests all still pass and I've also tested it in an existing app and a new app without any issues.

I'm hoping you agree that this will make the use of this addon more straightforward and reduces the chances of issues due to the styling pipeline.


tests/dummy/app/styles/_variables.css
```diff
@@ -0,0 +1,34 @@
+/* Colors */
+
+:root {
```
maxfierke: I believe CSS variables here without PostCSS might kinda break the demo site on IE11, but I think that's probably fine.

ember-cli-build.js
```diff
@@ -37,5 +16,9 @@ module.exports = function(defaults) {
     behave. You most likely want to be modifying `./index.js` or app's build file
   */
 
+  app.import('vendor/highlight.pack.js', {
```
maxfierke: For posterity's sake, what were the options used for generating this? Might be worth adding a comment above.

tests/dummy/app/styles/_variables.css
chrism: I hope that the second commit should help with this — the variables now only override the default colour scheme so on IE it should display OK but without code highlighting.

chrism:
Thanks @maxfierke I've added the comment explaining the customization.

chrism:
@maxfierke thanks! well spotted 👍

addon/styles/addon.css
```diff
@@ -0,0 +1,145 @@
...
+}
+
+.ember-tooltip:after {
+  content: '';
+  position: absolute;
+  width: 10px;
+  height: 10px;
+  margin: -5px;
+  background: inherit;
+  transform: rotate(45deg);
```
sir-dunxalot: Main thing to be aware of with the removal of the `autoprefixer` dep is we drop support for a couple of [lesser-used old browsers](https://caniuse.com/#feat=transforms2d). To be extra safe, perhaps we can add in `-webkit` prefixes for `transform` and `transition` rules. However, I don't really mind either way since:

1) This is already supporting graceful fallbacks
2) We can deal with this if and when an issue is reported

So I'll leave the addition of prefixes up to @chrism's discretion.

tests/dummy/app/styles/_variables.css
sir-dunxalot: Am fine with using css variables. I'd be surprised if any devs are using IE11 😛

sir-dunxalot:
@maxfierke @chrism LGTM. 👍

addon/styles/addon.css
chrism: Hi @sir-dunxalot, thanks for the review.

My understanding is that `transform` & `transition` only remain unsupported by  [IE8 and opera mobile](http://doiuse.herokuapp.com/?url=&browsers=&css=transform%3A%20rotate(45deg)%3B).

Using a prefix (such as `-webkit`) wouldn't solve this for those two browsers (would require polyfilling I think) and as Ember 2.0 itself no longer supports IE8 I'd hoped it wouldn't be necessary to support it in this addon.

Also, another consideration is that as the `transform` is located in a pseudo element selector (`:after`) I don't think IE8 supports that anyway.

If I've got some of this wrong I'd be happy to update, but I don't think there is much possible for the reasons I've listed.

Thanks.

PR #265, opened 2018-09-09
Fix for glimmer 2 issue

kennydp:
This actually builds upon this open PR : https://github.com/sir-dunxalot/ember-tooltips/pull/176. 

I just added what was suggested. If you are using the wrapper, the target is the grandparent element. If you are not using it, it will still be the same code as before. This should not break anything for people using the thether-tooltip-and-popover by itself.

This should fix the glimmer 2 exceptions people are having when using this addon (https://github.com/sir-dunxalot/ember-tooltips/issues/122).

### DIFF
--- a/addon/components/lazy-render-wrapper.js
+++ b/addon/components/lazy-render-wrapper.js
@@ -90,6 +90,9 @@ export default Component.extend({
   _hasUserInteracted: false,
   _hasRendered: false,
   _shouldShowOnRender: false,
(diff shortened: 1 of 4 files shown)


maxfierke:
Hi @kennydp! Thanks for the PR. Could you add a test case for the bug that this is fixing?

kennydp:
Sure thing. I'll try to add one when I find some time later this week

PR #267, opened 2018-09-11
fix: run handlers within the runloop

alexlafroscia:
Replaces my previous PR #240 

This also aims to accomplish the same thing as #199 -- maybe that PR should be closed as well?


alexlafroscia:
Addressed the cases in those files!

alexlafroscia:
Just a note: I'm not sure about your thoughts on Prettier, but if you'd be interested I'd love to add it here. It would make it quite a bit easier to contribute since the formatting throughout these files is sometimes a bit strange/inconsistent!

PR #268, opened 2018-09-13
Prevent layout thrashing in popper onUpdate hook

### LINKED ISSUE #236: With the 3.0.0-beta.x code, page scrolling FPS slows down after hiding tooltip (60 goes down to 30 at best, 15 at worst)
We'e got a page where the tooltip is shown based on a dynamic property, and hidden 1.5 seconds later using an `Ember.later` call.

To get a baseline for scrolling FPS without the tooltip, I refresh the page, start Chrome's profiling tools, and scroll up and down continuously (using the mouse wheel) for 5 seconds before stopping the profiler. This way, I get a consistent 57-60 FPS the whole time.

To test with the tooltip, I refresh the page again, show the tooltip, and wait for it to hide before starting the profiler and repeating the test.

This time, I can see a visible jank almost immediately. Stopping the profiler after 5 seconds shows me roughly 60 FPS for the first 500ms or so, and then it drops to anywhere between 15 and 30 FPS (although it looks really, really jittery). If I leave the profiler recording for 10 or 15 seconds, sometimes it'll crash and take the whole tab down with it.

According to the profiler, the culprit seems to be the vast amount of times that Popper is running its `requestAnimationFrame` call: 
```
this.scheduleUpdate = function () {
    return requestAnimationFrame(_this.update);
};
```

This call is not run at all before the tooltip is shown, so does it really need to be shown after the tooltip is shown and then hidden?

Granted, it might need to run while the tooltip is shown, to reposition it... but for the massive slowdown it causes, can the positioning calculations be disabled when the tooltip is hidden, and run again when shown?

Thanks!

maxfierke:
To avoid thrashing layout, we can wrap the style accessing in a `requestAnimationFrame` to ensure changes to spacing are animated smoothly, and we can add some guards to avoid computing spacing for destroyed or hidden tooltips.

Fixes #236

### DIFF
--- a/addon/components/ember-tooltip-base.js
+++ b/addon/components/ember-tooltip-base.js
@@ -372,19 +372,31 @@ export default Component.extend({
   },
 
   setSpacing() {
-    const { popperInstance } = this.get('_tooltip');
-    const { popper } = popperInstance;
-    const marginSide = getOppositeSide(popper.getAttribute('x-placement'));
-    const { style } = popper;
+    if (this._spacingRequestId) {
+      return;
+    }
 
-    style.marginTop = 0;
-    style.marginRight = 0;
-    style.marginBottom = 0;
-    style.marginLeft = 0;
+    this._spacingRequestId = requestAnimationFrame(() => {
+      this._spacingRequestId = null;
 
-    popper.style[`margin${capitalize(marginSide)}`] = `${this.get('spacing')}px`;
+      if (!this.get('isShown') || this.get('isDestroying')) {
+        return;
+      }
 
-    popperInstance.state.updateBound();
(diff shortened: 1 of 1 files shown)


sir-dunxalot:
@maxfierke This PR is good to go for the next 3.x beta release, right?

maxfierke:
@sir-dunxalot yes, I think so. There are still some framerste dips going on, but it seems to fix the 100% CPU issue.

sir-dunxalot:
Ok. That's a good step forward even if it's not perfect! If you think we should return to the `framerate` issue, just open a new issue as a reminder. I trust your judgement either way since I haven't recently looked into the performance issues at all.

PR #269, opened 2018-09-21
Ember 3 test syntax

sir-dunxalot:
- Updates the test suite to use Ember 3. Examples of the new test syntax can be found in [this blog post](https://dockyard.com/blog/2018/03/29/testing-your-ember-application-in-2018).
- Test helpers have been updated to support `async` functions whilst remaining backwards-compatible


package.json
```diff
@@ -47,6 +47,7 @@
     "ember-cli-shims": "^1.2.0",
     "ember-cli-sri": "^2.1.0",
     "ember-cli-uglify": "^2.0.0",
+    "ember-cli-update": "^0.27.0",
```
maxfierke: We should probably omit this and use it globally, since we'll want to be using whatever the latest is, when we run it.

package-lock.json
```diff
@@ -0,0 +1,14761 @@
+{
```
maxfierke: what do you think about ignoring `package-lock.json` and `yarn.lock` in the `.gitignore`. It seems like it may not make sense to commit these, since it won't matter for runtime users of the addon.

tests/acceptance/acceptance-test.js
```diff
@@ -1,88 +1,67 @@
...
+    assertTooltipRendered(assert, popoverOptions);
 
-    assertTooltipVisible(assert, options);
+    assertTooltipVisible(assert, popoverOptions);
 
-    triggerTooltipTargetEvent($popoverTarget, 'mouseleave');
+    await triggerTooltipTargetEvent($popoverTarget, 'mouseleave');
 
-    afterTooltipRenderChange(assert, () => {
-      assertTooltipNotVisible(assert, options);
```
maxfierke: should this assertion be preserved?

sir-dunxalot: Yes, this looks like an error. I will add the assertion back.

sir-dunxalot:
@maxfierke Tests are now passing and comments have been addressed. Let me know if you want more time to review the PR or just give the thumbs up and I will merge.

PR #280, opened 2018-10-03
Fix compatibility breaks in 3.x tooltip test helpers

maxfierke:
Restores `targetSelector` option, which is useful for targeting based
a specific tooltip out of many when tooltips share a common class, but
have differentiatable target elements.

Also, restores the ability to target based on the `class` passed to
`{{ember-tooltip}}`.

With these changes in place, host app test suites should be able to migrate over from 2.x fairly cleanly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -799,7 +799,7 @@ test('Example test', async function(assert) {
   await triggerEvent(this, this.element);
 
   assertTooltipVisible(assert, {
-    selector: '.target-b', // Or whatever class you added to the target element
+    targetSelector: '.target-b', // Or whatever class you added to the target element
   });
 });
 ```
--- a/README.md
+++ b/README.md
@@ -389,9 +389,6 @@ Four actions are available for you to hook onto through the tooltip/popover life
 }}
 ```
 
-As of 2.11.0, specifying just the name of an action (e.g. `'onDestroy'`) is
-deprecated, and will be removed in 3.0.0. Please use closure actions instead.
-
 ## Testing
 
 ### Test helpers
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -761,6 +768,42 @@ test('Example test', async function(assert) {
...
+    </div>
+    <div class="target-b">
+      {{ember-tooltip class="common-tooltip" side='left' isShown=true text='Bye' effect='none'}}
+    </div>
+  `);
+
+  await triggerEvent(this, this.element);
+
+  assertTooltipVisible(assert, {
+    selector: '.target-b', // Or whatever class you added to the target element
```
sir-dunxalot: Is this example for `selector` or `targetSelector`? It should be the latter.

addon-test-support/utils/find-tooltip.js
```diff
@@ -11,7 +11,12 @@ export default function findTooltip(selector) {
   children of <body> instead of children of the $targetElement */
 
   const $body = $(document.body);
-  const $tooltip = $body.find(selector);
+  let $tooltip = $body.find(selector);
+
+  if ($tooltip.length && $tooltip.hasClass('ember-tooltip-base')) {
```
sir-dunxalot: Probably worth documenting why this `if` statement is necessary.

README.md
maxfierke: doh! yup, good catch!

addon-test-support/utils/find-tooltip.js
maxfierke: good call. will do 👍

Issue #279, opened 2018-09-28
Tooltip hidden when inside a container with overflow: auto


transitorytammy:
For embertooltips 3.0, since the tooltip is now a sibling to the element, there is an issue when a parent container has `overflow: auto/scroll`, the tooltip sometimes gets cut off.

<img width="385" alt="screen shot 2018-09-28 at 2 24 51 pm" src="https://user-images.githubusercontent.com/1967604/46234316-bbb45300-c32a-11e8-9ec1-ca1db230d90d.png">

Is there an option to show the tooltip in this scenario?

maxfierke:
@transitorytammy you may want to experiment with passing a `targetId` to `{{ember-tooltip}}`, which will allow you to target a different parent element

Let me know if that works for your use case, as this came up with the travis-ci folks recently too, so it might be good for us to document in the coming upgrade documentation!

transitorytammy:
hey @maxfierke!  I did try doing that but it doesn't work in my case.  It might be because there are a few levels of divs between the parent with the `overflow-y: auto` and where the tooltip is.

maxfierke:
Interesting... would you be able to put together a small example that reproduces your use case's DOM structure? (doesn't need to be a live example, just a snippet that I could use to write a test or build out a new ember app with to play around)

transitorytammy:
So, the DOM structure is basically like this.

```
<div class="outerContainer">
  <table>
    {{#each sortedTerms as |keyword index|}}
      <tr class="row">
        <td class="term">
          {{#ember-tooltip}}
            {{keyword.term.hint}}
          {{/ember-tooltip}}
          <span>{{keyword.text}}</span>
        </td>
        <td class="strength">
          {{#ember-tooltip}}
            {{keyword.term.hint}}
          {{/ember-tooltip}}
          <span>{{keyword.text}}</span>
        </td>
      </tr>
  </table>
</div>
```

and `.outerContainer` has `overflow-y: auto` on it that is causing the issues for me.

maxfierke:
@transitorytammy I played around with this a bit. It seems to work as expected if `position` is unset or `static`, but if `.outerContainer` is `position: relative` or `position: absolute` it exhibits this behavior. Is `.outerContainer` or another parent of the tooltip positioned this way by any chance? 

If so, I was able to resolve that behavior by removing the `preventOverflow` option we pass into popper.js (added in 3.0.0-beta.3). We're hard-coding the popper options right now

[...]

Issue #275, opened 2018-09-24
3.0.0-beta.3: some tooltips don’t render or render strangely


backspace:
Hey all, thanks for the renewed push of work on this addon, it’s quite useful for us in `travis-web`! I was trying out the new beta as suggested in Discord and I found at least one set of tooltips that doesn’t render at all, or occasionally renders in a distorted fashion, maybe related to proximity to the viewport edge. An example, which you can try out [here](https://bd-tooltips-3.test-deployments.travis-ci.org/dashboard/builds) if you log in and turn on the “Dashboard” feature flag in your account settings:

![missing-tooltip](https://user-images.githubusercontent.com/43280/45966184-06af2d00-bff0-11e8-958a-2bd5db34b167.gif)

If you look at the [template](https://github.com/travis-ci/travis-web/pull/1813/files#diff-2ef062b5530a7b1795e8ebc728df1bf4), there’s no obvious difference between the invocation of the tooltip that works and the one that doesn’t. Strangeness!

Let me know if I can help in any way to figure this out. 💞

maxfierke:
Hi @backspace, thanks for the report! My first blink is that it might be the same issue or related issue to #225. I'm going to take a look at fixing that tonight, so hopefully I'll know more then!

backspace:
oh wow, thanks for figuring that out, I’m sorry to have misdirected responsibility! 😳

maxfierke:
All good! We'll be sure to document the CSS changes & potential gotchas when writing up the 2.x to 3.x migration docs.

maxfierke:
@backspace 3.0.0-beta.5 (now latest) adds support for passing `popperOptions`. I think something like this _may_ fix the behavior you're seeing without needing a CSS workaround:

`{{ember-tooltips text=tooltipText popperOptions=popperOptions}}`

where `popperOptions` is set to something like this:

```json
{
  "modifiers": {
    "preventOverflow": {
      "escapeWithReference": false
    }
  }
}
```

backspace:
Thanks for the tip! I actually wasn’t able to figure out a good CSS fix for this as there’s some weirdness to add fading for very long strings, so I hope to find another way.

I tried this locally with beta 5 and it didn’t seem to help; I can only see a hint of the tooltip while scrolling:

![tooltip-overflow](https://user-images.githubusercontent.com/43280/47037764-7030e000-d145-11e8-9c81-70c2f18016b7.gif)

[...]

PR #285, opened 2018-10-05
Add support for overriding popperOptions

maxfierke:
Allows users to supply their own `popperOptions` to the underlying `tooltip.js` instance merging with the defaults set by `ember-tooltips`. This will allow doing things like disabling `popperOptions.modifiers.preventOverflow.escapeWithReference` for situations such as #275 & #279, and other use-cases where folks need some fine-tuning of the how popper is rendering the tooltip.

### DIFF
--- a/addon/components/ember-tooltip-base.js
+++ b/addon/components/ember-tooltip-base.js
@@ -3,6 +3,7 @@
 import $ from 'jquery';
 import { computed } from '@ember/object';
 import { getOwner } from '@ember/application';
+import { assign } from '@ember/polyfills';
 import { run } from '@ember/runloop';
 import { warn } from '@ember/debug';
 import Component from '@ember/component';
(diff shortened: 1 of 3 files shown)


addon/components/ember-tooltip-base.js
```diff
@@ -554,3 +559,27 @@ export default Component.extend({
     }
   }
 });
+
+function mergeModifiers(defaults, overrides = {}) {
```
sir-dunxalot: Could we use [Ember's `merge()` method](https://emberjs.com/api/api/ember/release/functions/@ember%2Fpolyfills/merge) instead?

sir-dunxalot: Or failing that, using the spread operator [detailed here](https://davidwalsh.name/merge-objects)?

maxfierke: `Ember.merge` is deprecated in ember-beta (3.5, I believe). The problem with both is that they don't do deep merges, which is what we'd need to preserve defaults that are not overridden. This is a pretty naïve deep merge implementation (only depth=1), but we can totally refactor it and/or swap in a good deep merge library later, if our needs change.

Issue #238, opened 2018-02-20
Uncaught TypeError: Cannot read property 'length' of null


vastan:
The error from title appears when I'm trying to add 2 elements that contain tooltips after another ember addon called ember-line-clamp. The error is shown only on multiple elements after the line-clamp addon. It seems to work just fine with only one element to contain the tooltip.

Version 3.0.0 beta 2

Here you can find an example: https://github.com/mihnea-baia/test-ember-tooltips

sir-dunxalot:
Thank you for the code! Very helpful. I can confirm I am seeing the error when I run mihnea-baia/test-ember-tooltips locally and the error is gone when removing either the `{{line-clamp}}` component or the `{{ember-tooltip}}` component.

I will investigate the source of the error and report back, hopefully with a PR that fixes this!

sir-dunxalot:
This issue has been fixed and is not present with the latest ember-tooltips prerelease, `3.0.0-beta.4`, which you can install using `ember install ember-tooltips@3.0.0-beta.4`.

To be honest, I don't know exactly what fixed this issue and, since everything is green in the latest release, I didn't dig in any further. #286 adds a test to monitor future compatibility with `ember-line-clamp`.

sir-dunxalot:
Fixed by #272.

PR #286, opened 2018-10-07
Added test to monitor compatibility with ember-line-clamp

sir-dunxalot:
Adds a test to make sure we don't regress to #238, which is inadvertently fixed in the latest ember-tooltips prerelease, `3.0.0-beta.4`.


sir-dunxalot:
@maxfierke Would like a second opinion on whether this is a necessary test to add given #238. What do you think?

sir-dunxalot:
My best guess as to what fixed it was one of the commits related to `rAF` and the run loop since both libraries rely quite a bit on `rAF`. Perhaps f4193439bf7da5f13b06f2547ce63824fa9f8a05 or e10ea09bb2e1f7e00288e7575e8e2b8ac53b9746.

sir-dunxalot:
@maxfierke `yarn.lock` has been deleted. I'd like to use `yarn.lock` instead of `package.json` but I don't want to add noise to the PR. I will handle that change in a separate, future PR tracked by #288.

PR #296, opened 2018-10-18
Popover parity

### LINKED ISSUE #295: isShown cannot be used to hide() popover
When isShown changes to true, the popover displays.
When isShown changes back to false, hide() doesn't dismiss the popover. 

Do we know why this line of code was added? 
https://github.com/sir-dunxalot/ember-tooltips/blob/3.x/addon/components/ember-popover.js#L99
hide() is essentially a noop when isShown === false

sir-dunxalot:
- Adds `isShown` test, which fixes #295
- Removes `ember-getowner-polyfill is not required for Ember 2.3.0 and later, please remove from your package.json` warning


addon/templates/components/lazy-render-wrapper.hbs
```diff
@@ -1,23 +0,0 @@
-{{#if _shouldRender}}
```
basicBrogrammer: hahahah I was wondering about this

sir-dunxalot:
The push is passing but the PR isn't. It seems like an unrelated issue (a single test is failing for the `ember-release` scenario). I'm going to merge into `3.x` and hopefully that will solve the odd test suite failure.

Issue #304, opened 2018-10-31
Question: how to render tooltip on body


timmorey:
I'm using 3.0, and finding that the tooltip gets rendered in-place, rather than as a direct child of the application's root element.  This is leading to some layout issues for us.

I see that I can set the `APP.rootElement` in my `config/environment.js` to override this (https://github.com/sir-dunxalot/ember-tooltips/blob/master/addon/components/ember-tooltip-base.js#L317-L326).  Currently, I am not setting this, as I want to use ember's default behavior, which roots the application on the `body`.  It seems wrong to set this configuration property just for this addon.

I'm wondering if the `ember-tooltip-base` component should default the `container` to the body element, rather than `false`, if the config property is not set.

maxfierke:
hi @timmorey, you may want to try passing `targetId` where `targetId` is the ID of an element you'd like to be the tooltip's parent. For instance, you might try something like

```html
<body>
  <div id="number-one-tooltip-zone" class="tooltip-zone"></div>
</body>
```

```hbs
{{ember-tooltip text="Hello" targetId="number-one-tooltip-zone"}}
```

timmorey:
Thanks for the response @maxfierke.  It appears that setting the `targetId` controls what the tooltip points at -- this is not what I'm looking for.  Let me try to clarify.

When I add an `{{ember-tooltip}}`, it points at its parent by default -- this is the behavior I want.   I'm looking for a way to have the tooltip rendered at a higher level in the dom.  Perhaps an illustration will clarify:

```
<body class='ember-application'>
  <!-- I expect the tooltip dom to be rendered here -->

  ...
    <div class='deeply-nested-div-that-needs-explanation' >
      {{ember-tooltip text='explain that div'}}
    </div>
</body>
```

This was the default behavior in 2.x of this addon - what is the correct way to achieve this with 3.x?

maxfierke:
hah, I think we're talking about the same thing @timmorey. My misunderstanding was around how the `targetId` was getting used by popper. Agreed, we should definitely support this. Your blink on providing a way to override the `container` property in popper is correct.

maxfierke:
@sir-dunxalot what do you think about changing the default to `'body'` rather than `false`, to preserve the expectation from 2.x?

timmorey:
[...]

Issue #305, opened 2018-11-01
Popover click-to-hide test is flaky


maxfierke:
Every so often, there's a build failure on CI for the `Popover: click target, click hide-action, click target` test. It often passes when retried.

```
not ok 67 Chrome 70.0 - [665 ms] - Integration | Option | API: Popover: click target, click hide-action, click target
    ---
        actual: >
            false
        expected: >
            true
        stack: >
                at assertTooltipVisible (http://localhost:7357/assets/test-support.js:14435:12)
                at Object._callee2$ (http://localhost:7357/assets/tests.js:2662:55)
                at tryCatch (http://localhost:7357/assets/vendor.js:404:40)
                at Generator.invoke [as _invoke] (http://localhost:7357/assets/vendor.js:678:22)
                at Generator.prototype.(anonymous function) [as next] (http://localhost:7357/assets/vendor.js:437:21)
                at step (http://localhost:7357/assets/tests.js:2543:32)
        message: >
            assertTooltipVisible(): the ember-tooltip should be visible:
                  aria-hidden = true
        negative: >
            false
        Log: |
```

PR #308, opened 2018-11-05
Add container option

timmorey:
Proposed solution for https://github.com/sir-dunxalot/ember-tooltips/issues/304

Adds a `container` attribute to the `ember-tooltip` component, which is passed down as [tooltip.js' container option](https://popper.js.org/tooltip-documentation.html#new_Tooltip_new).  It defaults to `false`, which will render the tooltip as a sibling of its target.  Optionally, it can be set to an element or selector that identifies an alternate container for the tooltip.

In this changeset, the `APP.rootElement` config property is ignored - it used to be used to populate the container.  The previous behavior prevented apps that have this config property set from being able to render tooltips in-place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ For example, if you want to show a tooltip over a button when the user hovers ov
 Options are set as attributes on the tooltip/popover components. Current tooltip/popover properties this addon supports are:
 
 - [class](#class)
+- [container](#container)
 - [delay](#delay)
 - [delayOnChange](#delay-on-change)
 - [duration](#duration)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -149,6 +150,20 @@ Adds a class to any tooltip:
 {{ember-tooltip class='tooltip-warning'}}
 ```
 
+#### Container
+
+| Type | HTMLElement \| String \| false |
```
maxfierke: I think these backslashes are messing with the formatting of the headers in the rendered markdown

maxfierke: Oooooooo hah no, I was just being bad at reading. That makes sense. Making them appear as code blocks would be more clear though.

timmorey:
@maxfierke any idea what is causing the CI failure: https://travis-ci.org/sir-dunxalot/ember-tooltips/jobs/451110952#L610?  I seem to be getting occasional/inconsistent failures from that test, and it appears to be unrelated to these changes.

maxfierke:
@timmorey that's #305. I'm hoping to take a look at it some time this week. It's flaky and should pass eventually :/

Issue #277, opened 2018-09-27
Popper.js sourcemap error on 3.0.0-beta.3


shanicem:
I'm trying to upgrade to the 3.0.0-beta.3 version but I get this error when I try to start the server:

`ENOENT: no such file or directory, open '/Users/username/project-name/tmp/source_map_concat-input_base_path-N0GNP8fr.tmp/vendor/popper/popper.js'`

It seems similar to issue #226 

Is this still a known issue?

maxfierke:
@shanicem Thanks for the report! This instance seems a little different, since it can't seem to find `popper.js` rather than `popper.js.map`. Could you post which version of ember-cli you're using and which version of `popper.js` npm has installed? (I think `npm list popper.js` should show the version)

shanicem:
@maxfierke thanks for the quick reply! we're using `popper.js@1.14.4` and `ember-cli@3.3.0`

maxfierke:
@shanicem Unfortunately, I was unable to reproduce this with a newly generated ember 3.3 app, which suggests to me that this might be environmental. I'd suggest clearing `node_modules`, `tmp` and `dist`, and doing an `npm install` from scratch to see if that fixes it.

If you're still seeing the error after that, I'd be curious about what other addons your app is using. Perhaps there's a conflict somewhere!

shanicem:
@maxfierke Thanks for your response. We tried what you suggested above, as well as upgrading to beta.4, but still getting the same issue. We're just going to move to 2.11.1 for now which is working for us. Thanks again for all your help!

sir-dunxalot:
@shanicem This seems like a rare but serious error. Even though `2.11.1` is working for you, it would be very helpful if you could provide some information that @maxfierke or I can use to debug this issue.

Are you able to share your `config/environment.js` and  `ember-cli-build.js` files? Alternatively, if you are able to give access to a repo that has a branch demonstrating this issue, we can look at the issue there.

shanicem:
@sir-dunxalot Unfortunately I can't share them publicly. Would it be possible to send them to you in an e-mail?

sir-dunxalot:
@shanicem, yes of course. I just sent you an email that you can respond to.

sir-dunxalot:
I am also going to investigate whether this issue appears when `popper.js` or `ember-popper` are installed. For example, it has been suggested that this issue is present when `ember-bootstrap` is installed, which uses `ember-popper`.

[...]

PR #311, opened 2018-11-06
Remove ember-cli-node-assets

camerondubas:
Hey there 👋

This removes the use of `ember-cli-node-assets` which is, in the words of maintainer, "[mostly obsolete](https://github.com/dfreeman/ember-cli-node-assets#ember-cli-node-assets--)" in favour of importing from `node_modules` directly. 

I'm not sure which versions of Ember this addon wishes to support, but directly importing node_modules was added in 2.15, so this will break in versions earlier than that.

This fixes https://github.com/sir-dunxalot/ember-tooltips/issues/277 at least in that specific case (I work with @shanicem, who opened that issue)

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,32 +3,10 @@
 module.exports = {
   name: require('./package').name,
 
-  options: {
-    nodeAssets: {
-
-      'popper.js': {
-        vendor: {
-          srcDir: 'dist/umd',
-          destDir: 'popper',
-          include: ['popper.js', 'popper.js.map'],
-        },
-      },
-
-      'tooltip.js': {
-        vendor: {
-          srcDir: 'dist/umd',
-          destDir: 'popper',
-          include: ['tooltip.js', 'tooltip.js.map'],
-        },
-      },
-    },
-  },
-
   included: function(app) {
     this._super.included(app);
 
-    app.import('vendor/popper/popper.js');
-    app.import('vendor/popper/tooltip.js');
-  },
-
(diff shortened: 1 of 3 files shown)


maxfierke:
Very cool @camerondubas! I'm going to poke around a bit to see how this works with engines and when `ember-tooltips` is included as a dependency in another add-on to make sure the popper dependency gets imported correctly in those cases too.

Last @sir-dunxalot and I talked about ember-cli version support, I believe we settled on supporting as far back as 2.12 (it was still included in the default Travis matrix from Ember-CLI then.) However, we might be able to use `ember-cli-version-checker` to fallback to the `ember-cli-node-assets` approach on older versions, and use this newer approach on newer ember-cli versions.

camerondubas:
Sounds good, let me know if there's anything I can do to help.  I can implement `ember-cli-version-checker` and add it to this PR if you'd like.

Also, we're currently using my fork of `ember-tooltips` which uses this fix in production, and we have a lazy-loaded engine that makes use of it without any issues.

sir-dunxalot:
@maxfierke I removed 2.12 from the `ember-try` matrix in a recent PR. The earliest version of Ember that `3.x`  tests for is `2.16` ([test matrix here](https://github.com/sir-dunxalot/ember-tooltips/blob/master/config/ember-try.js)).

I defer to @maxfierke for this PR, including investigating compatibility with engines (is it possible to write a test for that?), but in concept this is a good improvement to `ember-tooltips`.

PR #317, opened 2018-11-13
Fix flaky test by mimicking mouse events

### LINKED ISSUE #305: Popover click-to-hide test is flaky

maxfierke:
The click on element and click on hide-action both simulate the related mouse events, which have an impact on how the popover is shown and hidden (popover will avoid hiding if mouse is present.) Therefore, we simulate leaving the element so that it hides every time!

(* I'm 96% sure that's what was happening here.)

Fixes #305

### DIFF
--- a/tests/integration/components/popover/api-test.js
+++ b/tests/integration/components/popover/api-test.js
@@ -51,11 +51,12 @@ module('Integration | Option | API', function(hooks) {
 
     assertTooltipNotRendered(assert);
 
-    await triggerEvent(element, 'click');
+    await click(element);
 
     assertTooltipVisible(assert);
 
     await click('.hide-action');
+    await triggerEvent(element, "mouseleave");
 
     assertTooltipNotVisible(assert);
 


sir-dunxalot:
You are a god! Hopefully this is an easy fix to a very annoying bug. Even if you have uncertainty whether this 100% fixes the issue, I think it's worth merging and observing.

Issue #309, opened 2018-11-05
ember-wormhole error


transitorytammy:
Hi hi. 

I am getting this error randomly in production from `ember-wormhole` which is a dependency of `ember-tooltip`.  

`The head node of a wormhole must be attached to the DOM`

https://github.com/yapplabs/ember-wormhole/blob/master/addon/components/ember-wormhole.js#L63

Wondering if you had suggestions on the best way to solve this on my side?

maxfierke:
Hi @transitorytammy! Which version(s) of ember-wormhole is resolved & installed in your app? Could you also share a sample of how `ember-tooltips` is being used when you get this error and the full stack trace, if possible?

timmorey:
I've been running into this one too, but haven't managed to repro outside of my app.  It appears to be a timing issue, where an `ember-tooltip` gets rendered and then removed before its wormhole is fully initialized.  In particular, I'm seeing https://github.com/yapplabs/ember-wormhole/blob/master/addon/components/ember-wormhole.js#L74 run immediately before https://github.com/yapplabs/ember-wormhole/blob/master/addon/components/ember-wormhole.js#L60-L65.

So, there is some questionable behavior in my app, which is causing content to be rendered and then immediately replaced, though it would still be nice if we could avoid the crash.

I'm running 
ember-tooltips 3.0.1
ember-wormhole 0.5.5
ember-source 3.0.0 (working on upgrading to a more recent version)

More complete error (stack trace truncated, since it is incredibly long)

```
Uncaught Error: The head node of a wormhole must be attached to the DOM
    at run.schedule (ember-wormhole.js:71)
    at invoke (backburner.js:205)
    at Queue.flush (backburner.js:125)
    at DeferredActionQueues.flush (backburner.js:278)
    at Backburner.end (backburner.js:410)
    at Backburner._boundAutorunEnd (backburner.js:372)
```

maxfierke:
@transitorytammy @timmorey Not sure if this is relevant or not: https://github.com/yapplabs/ember-wormhole/issues/71. It makes it sound like it could be nebulously related to an old `ember-cli-htmlbars` version or other dependencies, so if either of you are able to share with me your `package.json` and related lock file, that would help a lot!

Possibly a quick experiment: could you give it a shot upgrading (or try `3.0.2`, which bumped `ember-cli-htmlbars` requirement to `^3.0.0`)?

benclum11:
[...]

PR #319, opened 2018-11-17
Replace ember-wormhole with ember-maybe-in-element

maxfierke:
ember-maybe-in-element leverages the Glimmer native `-in-element`
helper to render content into another DOM node. This allows us to
drop ember-wormhole and may address the issues raised in #309.

### DIFF
--- a/addon/components/ember-tooltip-base.js
+++ b/addon/components/ember-tooltip-base.js
@@ -160,6 +160,14 @@ export default Component.extend({
     return `${this.get('elementId')}-wormhole`;
   }),
 
+  wormholeEl: computed('wormholeId', '_awaitingTooltipElementRendered', function() {
+    if (this.get('_awaitingTooltipElementRendered')) {
+      return null;
+    } else {
+      return document.getElementById(this.get('wormholeId'));
+    }
+  }),
+
   _awaitingTooltipElementRendered: true,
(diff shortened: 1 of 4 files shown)


addon/components/ember-tooltip-base.js
```diff
@@ -160,6 +160,14 @@ export default Component.extend({
     return `${this.get('elementId')}-wormhole`;
   }),
 
+  wormholeEl: computed('wormholeId', '_awaitingTooltipElementRendered', function() {
```
sir-dunxalot: Is `wormhole` a misnomer now?

maxfierke: Depends on how you look at it, I guess. Conceptually, it's still a bit of a wormhole.

maxfierke: An alternative would maybe be `portal`?  I think that's what React calls their wormholes.

joelluijmes:
Please see my comment on the issue https://github.com/sir-dunxalot/ember-tooltips/issues/309#issuecomment-441224463, this actually breaks the tooltip in my case.

maxfierke:
Going to close this for now. I think it's definitely something to look into later, but also might be better off implementing something ourselves with `-in-element` directly and skipping the dependency

PR #320, opened 2018-11-19
Remove unnecessary runloop and promise in createTooltip()

maxfierke:
Likely root cause of #309, as it was confirmed removing the outer `run` fixed the issue.


sir-dunxalot:
Technically this could be a breaking change if someone is calling `createTooltip().then(...)`. However, I'd be surprised if that is being done anywhere outside this library.

maxfierke:
@sir-dunxalot Agreed. They should be registering an `onRender` callback instead anyway, especially since this has never been documented as public API.

Issue #321, opened 2018-11-23
3.x moving your mouse over and out and over quickly will prevent visibility


arthur5005:
You can see it on the demo page with the first example.
http://sir-dunxalot.github.io/ember-tooltips/

![2018-11-23 15 43 36](https://user-images.githubusercontent.com/697846/48962601-a0cf1b00-ef36-11e8-892a-46fe34d588d7.gif)

arthur5005:
Found the problem. `_tooltip.hide()` doesn't get called synchronously, so it's possible to call `_tooltip.show()` twice before hide gets called (inside the `run.later`) I assume this is so we can wait for the 

https://github.com/sir-dunxalot/ember-tooltips/blob/3baca0509dba8a8f1264351f885240de24f2ee75/addon/components/ember-tooltip-base.js#L480-L504

It's Friday, and I already have a drink in my hand :D. But I imagine we could either:
a) Track when transitions are happening, then call hide/show synchronously in those cases
b) Track the `run.later`s on hide/show and cancel them whenever we hide/show

maxfierke:
I think you're right on the money there! It's on my to-do to add a bit more bookkeeping to our runloop scheduling, given a few recent runloop-related bugs.

Thanks for digging!

arthur5005:
Just added a PR as a suggestion. Works well on our fork.

PR #322, opened 2018-11-26
Fix issue hide/show race by cancelling the final hide callback on show

arthur5005:
Suggested fix for issue #321

Added a `_completeHideTimer` to track the final `run.later` which actually completes the hiding of the tooltip. This is different than the `_hideTimer` which tracks a desired hide of the tooltip based on a duration supplied by the user.

This timer gets cancelled on every show to prevent a race condition between showing and hiding the tooltip.

### DIFF
--- a/addon/components/ember-tooltip-base.js
+++ b/addon/components/ember-tooltip-base.js
@@ -423,6 +423,7 @@ export default Component.extend({
     const duration = this.get('duration');
 
     run.cancel(this.get('_showTimer'));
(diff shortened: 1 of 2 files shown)


arthur5005:
@maxfierke Sounds good. I threw one in, it fails on master and passes on this one. Not sure if I stuck it in the right place. Let me know.

arthur5005:
Oops, had a couple of ESLint problems. Should be good now.

PR #325, opened 2018-12-05
Fix keydown bubbling

rob-long:
The keydown event handler for the escape key stops bubbling/propagation on any esc press.  If there are multiple tooltips loaded then the first tooltip to respond will disable all other tooltips from responding.  

Moreover, ember-tooltips is preventing other listeners from other modules or components from responding as well.

This branch fixes that by only stopping propagation when we know that a tooltip has been hidden.

### DIFF
--- a/addon/components/ember-tooltip-base.js
+++ b/addon/components/ember-tooltip-base.js
@@ -293,7 +293,8 @@ export default Component.extend({
     this._addEventListener('keydown', (keyEvent) => {
 
       if (keyEvent.which === 27) {
-        if (this.hide()) {
+        if (this.get('isShown')) {
+          this.hide();
           keyEvent.stopImmediatePropagation(); /* So this callback only fires once per keydown */
           keyEvent.preventDefault();
           return false;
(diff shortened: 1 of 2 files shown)


addon/components/ember-tooltip-base.js
```diff
@@ -292,14 +292,12 @@ export default Component.extend({
...
 
-      keyEvent.stopImmediatePropagation(); /* So this callback only fires once per keydown */
-
       if (keyEvent.which === 27) {
-        this.hide();
-
-        keyEvent.preventDefault();
-
-        return false;
+        if (this.hide()) {
```
maxfierke: `this.hide()` has async side-effects, so I don't think it's really suited to returning a value and being used in this way. Checking `this.get('isShown')` would be better, I think.

rob-long: Updated!

rob-long:
Added tests!

tests/integration/components/keydown-test.js
```diff
@@ -0,0 +1,45 @@
+//import { later } from '@ember/runloop';
```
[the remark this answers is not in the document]
rob-long: done!

PR #327, opened 2018-12-07
Allow base class to be customised

### LINKED ISSUE #324: Allow custom template HTML
Currently, the tooltip wrapper is, (as far as I can see,) uneditable:
https://github.com/sir-dunxalot/ember-tooltips/blob/8983cd7f8d90d09510aeb94d24fadbcc1bfc7876/addon/components/ember-tooltip-base.js#L321-L324

This is causing issues with our repo on a few fronts:
 - as we also use `Bootstrap` tooltips in other parts of the app this is conflicting and causing CSS conflicts which are a bit of a pain to override and gain control of
 - the rest of our app uses BEM class names and I'm a little disappointed that we'd have to move away from this naming convention

Would it be possible to allow a custom template string to be passed in, keeping the current one as the default so as not to disrupt current users?

brookjordan:
Some code bases may have more than one type of tooltip and thus the class of `tooltip` is liable to collide.

We should allow the base class to be updated.

Closes #324

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ Adds extra classes to tooltip arrows.
 Usually used along with [`tooltipClass`](#tooltipclass).
 
 ```hbs
-{{ember-tooltip tooltipClass='hoverhelp__arrow'}}
+{{ember-tooltip arrowClass='hoverhelp__arrow'}}
 ```
 
 This will create html similar to:
--- a/README.md
+++ b/README.md
@@ -144,10 +144,47 @@ Options are set as attributes on the tooltip/popover components. Current tooltip
 |---------|---------|
 | Default | none    |
 
-Adds a class to any tooltip:
+Adds a class to any tooltip wrapper:
 
 ```hbs
-{{ember-tooltip class='tooltip-warning'}}
+{{ember-tooltip class='tooltip-wrapper'}}
+```
+**Note:** This is usually not what you want, as the wrapper itself is hidden by default. You are probably looking for [`tooltipBaseClass`](#tooltipbaseclass) or [`tooltipClassName`](#tooltipclassname).
+
+#### tooltipBaseClass
+
+| Type    | String          |
+|---------|-----------------|
+| Default | 'tooltip'       |
+
+Modifies the base class used for all tooltip, and extended upon by the tooltip content (`[tooltipBaseClass]-inner`) and the arrow (`[tooltipBaseClass]-arrow`).
+
+Useful to avoid conflicts with other libraries:
+
+```hbs
+{{ember-tooltip tooltipBaseClass='hoverhelp'}}
+```
+
+This will create markup similar to:
+```html
+<div class="hoverhelp">
+  <div class="hoverhelp-arrow"></div>
+  <div class="hoverhelp-inner"><!-- content --></div>
+</div>
+```
+
+#### tooltipClassName
+
+| Type    | String          |
+|---------|-----------------|
+| Default | 'ember-tooltip' |
+
+Adds classes to any tooltip.
+
+Useful for creating variations of tooltip:
+
+```hbs
+{{ember-tooltip tooltipClassName='tooltip-warning'}}
 ```
 
 #### Delay
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Options are set as attributes on the tooltip/popover components. Current tooltip
 - [spacing](#spacing)
 - [text (tooltip only)](#text)
 
-#### Class
+#### class
 
 | Type    | String  |
 |---------|---------|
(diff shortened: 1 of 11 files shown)


maxfierke:
@brookjordan 
> > I would advocate for passing those options and hardcoding them as `.ember-tooltip-arrow` and `.ember-tooltip-inner`, respectively
> 
> I'm a big fan of this, I'll update to make this happen.
> Oh, it appears that this class is already added for the arrow. I'm not sure that popper.js needs to be aware of the `inner` class – just the trigger and the arrow.
> 

Perhaps functionally it does not, but I still think we should be good consumers and provide them and  pass them in as the `arrowSelector` and `innerSelector` options to `Tooltip`

> > Second, having both `tooltipClassName` and `tooltipBaseClass` feels confusing to me
> 
> Also agreed. I did it this way as having `tooltipClassName` allows multiple classes to be passed, while `tooltipBaseClass` would only allow a single class as I'm using it to extend to the `` `${tooltipBaseClass}-tooltip` `` version.
> 
> Plus I wanted to keep this backwards compatible. I'm not sure if a feature like this warrants a major version bump.
> 

I've thought about this a bit. I think you're right about the compatibility, because `tooltipClassName` has been documented. I think what makes sense to do would be to call the new property `tooltipClass` (I want to avoid the phrase "base class") and deprecate `tooltipClassName` (more below).

> > just use the class passed to the component
> 
> I'd feel a little dirty hi-jacking this class. Especially as it may get `display: block;` applied to it, possibly then showing the actual container.

Hah, yeah, good point!

> > _tooltipVariantClass -- private property hardcoded to ember-tooltip for tooltips, and ember-popover for popovers. Applied to outermost div in template.
> 
> I'm a big fan of this. It's a pain having to pass in `ember-tooltip` into `tooltipClassName`. But, again, that would require a version bump.
> 

I think perhaps we do this, and then make `tooltipClassName` a [`deprecatingAlias`](https://emberjs.com/api/ember/3.5/functions/@ember%2Fobject%2Fcomputed/deprecatingAlias) for `_tooltipVariantClass`

> > * tooltipArrowClassName -- Defaulted to ''. Applied to arrow div.
> > * tooltipInnerClassName -- Defaulted to ''. Applied to inner div.
> 
> Also agreed, but I think I would still want to default these to:
> 
> ```js
> tooltipBaseClass:      'tooltip',
> tooltipArrowClassName: `${tooltipBaseClass}-arrow`,
> tooltipInnerClassName: `${tooltipBaseClass}-inner`,
> ```

Yes, these seem like sensible defaults.

maxfierke:
@brookjordan Friendly post-Holiday season ping 🔔 Let me know if you have any outstanding thoughts or questions on my suggestions!

brookjordan:
Sorry, this is still in my backlog.
I'll try to get back to you soon,

brookjordan:
@maxfierke 

> Perhaps functionally it does not, but I still think we should be good consumers and provide them and pass them in as the `arrowSelector` and `innerSelector` options to `Tooltip`
Ok, I dug deeper into the popper API. This makes sense. Especially if these classes are technically changeable.

> I've thought about this a bit. I think you're right about the compatibility, because `tooltipClassName` has been documented. I think what makes sense to do would be to call the new property `tooltipClass` (I want to avoid the phrase "base class") and deprecate `tooltipClassName` (more below).
If this is how we want to go forward, I'm happy with this.

Note that, despite using deprecatingAlias, this is still a breaking change.

Hopefully this fits what you're expecting.
Tests pass on my local, I'm not quite sure why Travis is upset. I'll take a look when I get another moment. This was more changed than I was expecting to need. Ha 😅

addon/components/ember-tooltip-base.js
```diff
@@ -238,14 +240,6 @@ export default Component.extend({
     this._dispatchAction('onDestroy', this);
   },
 
-  addTargetEventListeners() {
```
maxfierke: I'm thinking we should keep these and `hide()` defined in the base component, as people may be extending `EmberTooltipBase` and depending on those implementations.

addon/components/ember-tooltip-base.js
```diff
@@ -329,10 +328,12 @@ export default Component.extend({
       placement: this.get('side'),
       title: '<span></span>',
       trigger: 'manual',
-      template: `<div class="tooltip ${tooltipClassName} ember-tooltip-effect-${this.get('effect')}" role="tooltip" style="margin:0;margin-${getOppositeSide(this.get('side'))}:${this.get('spacing')}px;">
-                  <div class="tooltip-arrow ember-tooltip-arrow"></div>
-                  <div class="tooltip-inner" id="${this.get('wormholeId')}"></div>
-                  </div>`,
+      arrowSelector: `.${emberTooltipArrowClass}`,
```
maxfierke: If `emberTooltipArrayClass` is something like `ember-tooltip my-tooltip`, it should probably be turned into `.ember-tooltip.my-tooltip. Right now, it seems like it would end up being `.ember-tooltip my-tooltip`

addon/components/ember-tooltip-base.js
brookjordan: Fair enough.

addon/components/ember-tooltip-base.js
brookjordan: Good catch.

README.md
```diff
@@ -138,16 +138,83 @@ Options are set as attributes on the tooltip/popover components. Current tooltip
...
+Useful to avoid conflicts with other libraries:
+
+```hbs
+{{ember-tooltip tooltipClass='hoverhelp'}}
+```
+
+This will create markup similar to:
+```html
+<div class="hoverhelp">
+  <div class="tooltip-arrow"></div>
```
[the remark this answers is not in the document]
brookjordan: Ah, I was going to do that, but ended up leaving it as explicit class definitions for these.
What do you think? Shall I extend `hoverhelp` to `hoverhelp-arrow` and `hoverhelp-inner`? 
After some thinking, I decided maybe it's best to let the consumer decide, not making these decisions for them.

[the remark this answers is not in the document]
maxfierke: Nah, I think this is fine. The explicitness is probably better.

UPGRADING-3.x.md
```diff
@@ -37,24 +37,24 @@ e.g.
...
 -        enableLazyRendering=true
--        side='right'}}
+-        side="right"}}
 +      {{#ember-tooltip
 +        class="js-user-photo-tooltip"
-+        tooltipClassName="ember-tooltip user-banner__photo__tooltip"
-+        side='right'}}
++        tooltipClass="user-banner__photo__tooltip"
++        side="right"}}
          <img src={{user.photo_url}} alt="User photo" />
```
brookjordan: It's cute that we have alt tags even in tests ❤️

maxfierke: A11y is key! 😄

brookjordan:
@maxfierke I think this current state addresses your issues.

maxfierke:
@brookjordan I agree, there's just that one outstanding comment on the README. After that's addressed, I think it's probably all-set.

README.md
```diff
@@ -138,16 +138,83 @@ Options are set as attributes on the tooltip/popover components. Current tooltip
...
+| Type    | String          |
+|---------|-----------------|
+| Default | 'tooltip-arrow' |
+
+Modifies the base class used for all tooltip, and extended upon by the tooltip content (`[tooltipClass]-inner`) and the arrow (`[tooltipClass]-arrow`).
+
+Useful to avoid conflicts with other libraries:
+
+```hbs
+{{ember-tooltip tooltipClass='hoverhelp__arrow'}}
```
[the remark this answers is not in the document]
brookjordan: Agh, thought I'd fixed this.
🙈

maxfierke:
@brookjordan alright, lgtm! I’m traveling this week but I’ll find some time to merge this and cut a release. Thanks for all your work!

Issue #340, opened 2019-02-08
3.1.1 lazyRendering is not working anymore


GerritSommer:
I just updgraded from 2.10 to 3.1.1 and made all the changes as recommended.
I am heavily relying on lazyRendering as some tooltips access computed properties, which should only be computed when needed (-> when the popover is shown).

With the new version the content-block of the tooltip seems to be computed when the targetElement is inserted to the DOM.

I made a little twiddle (always wanted to say that), which shows the problem:
https://ember-twiddle.com/e26ba3e4b9bcd7377a757e9cd33b12c0?openFiles=templates.application.hbs%2Ctemplates.components.ember-tooltip.hbs

I am not sure if I missed some new convention on how to use this addon, but I'd appreciate some help.

maxfierke:
@GerritSommer I will refer you to this reply from an earlier issue for now (and I shall update the Upgrading documentation later to include this): https://github.com/sir-dunxalot/ember-tooltips/issues/330#issuecomment-448271612

Please let me know if you have any follow-up questions!

GerritSommer:
@maxfierke 	Oh great, thats exactly what I needed.
Thanks for the info and the addon.

Issue #330, opened 2018-12-18
Performance issue in 3.X


panthony:
On a given screen in my app I may have hundreds of possible popovers.

Some of them may be costly to render.

in 2.x I was using `enableLazyRendering` which was perfect since it did not render the content of the popover until it was actually displayed.

This option was removed in 3.x without, as far as I can tell, any replacement.

Unfortunately this make this addon unusable for me in its 3.x version.

panthony:
Turns out passing `popperContainer="body"` will make ember-tooltip behave like it was with `enableLazyRendering: true`.

Thus using this option will fix this issue in my case.

I'll leave it open unless you feel like this is an edge case.

maxfierke:
The use of `popper.js` in 3.x addresses performance in a couple different ways that mostly make the old lazy rendering option unnecessary.

1. It uses `requestAnimationFrame` to handle updates to the DOM, which provides smooth 60FPS updates.
2. It does not update or re-position tooltips that are not shown.
3. Tooltips are only rendered on activation & are torn down when hidden.

The content will still be rendered in the DOM, but is hidden and not rendered into a tooltip/popover until it's activated.

If the content inside the tooltip is what's costly to render, there is an escape hatch, e.g.:

```
{{#ember-popover as |popover|}}
  {{#if popover.isShown}}
      {{expensive-component-thing}}
  {{/if}}
{{/ember-popover}}
```

panthony:
Ok ! Good to know for `popover.isShown`.

And since the `popperContainer="body"` already workaround this issue for me (by only appending the tooltip when shown) i'll close this issue too.

Thanks for your time @maxfierke

PR #343, opened 2019-02-16
Avoid rendering content into DOM until popper element created

### LINKED ISSUE #341: Always wrapping tooltips in {{if
I think it would be very useful to have some property that if true, would render the tooltip, but if false, would not. I'm always having to wrap all of my dynamic data tooltips with an `{{if` check because otherwise it shows an empty tooltip, which is visually unappealing.

maxfierke:
Fixes #341 and more adequately addresses concerns from #340, #330 without requiring the user to check `isShown`. Bug was introduced by me in b83580c43689d922cb53e002479781332f9be4ef, and broke a promise about how enableLazyRendering in 2.x would essentially be the default and behave the same way.

However, the fastboot tests established that the content would be rendered into the DOM in fastboot mode, so this preserves that behavior.

This may break folks with existing tests depending on rendered content, so it seems like it may warrant a major version bump. But I can also see it being framed as a fix for a bug introduced in the betas, and warrant being a minor version bump. /cc @sir-dunxalot

### DIFF
--- a/addon/components/ember-tooltip-base.js
+++ b/addon/components/ember-tooltip-base.js
@@ -1,5 +1,6 @@
 import Tooltip from 'tooltip.js';
 import Ember from 'ember';
+import { getOwner } from '@ember/application';
 import { computed } from '@ember/object';
 import { assign } from '@ember/polyfills';
 import { run } from '@ember/runloop';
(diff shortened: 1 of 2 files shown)


addon/components/ember-tooltip-base.js
```diff
@@ -159,6 +160,19 @@ export default Component.extend({
     return `${this.get('elementId')}-wormhole`;
   }),
 
+  _fastboot: computed(function() {
```
sir-dunxalot: Any reason for not using `fastboot: service()` here (where service is `import { service } from '@ember/service';`)?

sir-dunxalot:
Regarding versioning, a minor release or patch seems fine here. The update is not in the public, documented API and applies mostly to tests. If you're especially concerned, I suggest a deprecation notice but IMO this update is fine as is (assuming all our tests are working correctly).

addon/components/ember-tooltip-base.js
maxfierke: If the `fastboot` service does not exist in the host app (i.e. in non-Fastboot apps), the injector will throw an error. This was the approach recommended by the [Addon author guide](https://www.ember-fastboot.com/docs/addon-author-guide#accessing-the-fastboot-service) for fastboot.

PR #347, opened 2019-03-05
Refactor jquery in tests

### LINKED ISSUE #307: Remove use of jQuery
jQuery is now an optional component of Ember. `ember-tooltips` has only one runtime use of jQuery that's easy to refactor out, but we should think about removing the need for jQuery within the test-helpers.

Perhaps we can continue to provide support for jQuery & jQuery selectors if jQuery is present, and use native DOM helpers otherwise.

mydea:
This PR changes the test helpers to work without jQuery.

In order to remain backwards compatible, all DOM-based test helpers were "copied" into a `dom` folder. You can import these variants like this:

```js
import { 
  findTooltip, 
  findTooltipTarget, 
  getPositionDifferences 
} from 'ember-tooltips/test-support/dom';
```

Also, the provided asserts use the new jQuery-free helpers under the hood - so if you use them, you don't need to change anything at all.

The new test helpers work mostly the same. However, `findTooltip` has a slightly different API, as it now needs to differentiate between one/multiple tooltips (vs. jQuery collections, which behave the same in both cases). 

`findTooltip` works in these ways:

```js
let tooltip = findTooltip(); // get first tooltip - returns a single DOM element
let allTooltips = findTooltip(null, { searchMultiple: true }); // get all tooltips - returns an array of DOM elements
let specificTooltip = findTooltip('.my-class');
let specificTooltips = findTooltip('.my-class', { searchMultiple: true });
let tooltipForTarget = findTooltip(myTargetDomElement);
```

Finally, it also now allows a `targetSelector` option, to get a specific tooltip element for a given target selector - which is used by `getTooltipAndTargetPosition()` under the hood.

Additionally, this PR also refactors all tests to stop using jQuery, and disables jQuery for the dummy app.

I guess in the future, the jQuery based test helpers should be deprecated, and finally removed.

This fixes #307 .

### DIFF
--- a/addon-test-support/dom/find-tooltip-target.js
+++ b/addon-test-support/dom/find-tooltip-target.js
@@ -1,0 +1,18 @@
+export function findTooltipTarget(selector) {
+  if (!selector) { // In case of passing null, undefined, etc
+    selector = '.ember-tooltip-target, .ember-popover-target';
+  }
+
+  const { body } = document;
+  const tooltipTarget = body.querySelectorAll(selector);
+
+  if (tooltipTarget.length === 0) {
+    throw new Error('ember-tooltips/test-support/dom/find-tooltip-target: No tooltip targets were found.');
+  } else if (tooltipTarget.length > 1) {
+    throw new Error('ember-tooltips/test-support/dom/find-tooltip-target: Multiple tooltip targets were found. Please provide an {option.targetSelector = ".specific-tooltip-target-class"}');
+  }
+
+  return tooltipTarget[0];
+}
+
+export default findTooltipTarget;
--- a/addon-test-support/dom/find-tooltip.js
+++ b/addon-test-support/dom/find-tooltip.js
@@ -1,0 +1,51 @@
+export function findTooltip(selectorOrElement, { targetSelector, multiple = false } = {}) {
+  if (!selectorOrElement) { // In case of passing null, undefined, etc
+    selectorOrElement = '.ember-tooltip, .ember-popover';
+  }
+
+  /* We querySelect tooltips from body instead of using ember-test-helper's
+  find() method because tooltips and popovers are often rendered as
+  children of <body> instead of children of the targetElement */
+
+  const { body } = document;
+  let tooltips = typeof selectorOrElement === 'string' ? body.querySelectorAll(selectorOrElement) : [selectorOrElement];
+
+  tooltips = [].slice.call(tooltips).map((tooltip) => getActualTooltip(tooltip, targetSelector)).filter((el) => el);
+
+  if (tooltips.length > 1) {
+    console.warn(`ember-tooltips/test-support/dom/find-tooltip: Multiple tooltips were found. Consider passing a selector '.specific-tooltip-class'`);
+  }
+
+  if (multiple) {
+    return tooltips;
+  }
+
+  let tooltip = tooltips[0];
+
+  if (tooltip && !tooltip.classList.contains('ember-tooltip') && !tooltip.classList.contains('ember-popover')) {
+    throw new Error(`getTooltipFromBody(): returned an element that is not a tooltip`);
+  }
+
+  return tooltip;
+}
+
+export default findTooltip;
+
+function getActualTooltip(tooltip, targetSelector) {
+  if (tooltip && tooltip.classList.contains('ember-tooltip-base')) {
+    /* If what we find is the actually the tooltip component's element, we can
+     * look up the intended tooltip by the element referenced by its target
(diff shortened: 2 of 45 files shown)


addon-test-support/dom/-get-closest-dom-parent.js
```diff
@@ -0,0 +1,22 @@
+/*
+  element.closest() is not supported in IE11, so for backwards compatibility we use this polyfill from
+  https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill
+ */
+export function getClosestDomParent(element, selector) {
```
maxfierke: I think this can _probably be omitted_ in favor of people providing the polyfill in their own test suites if they need it. I would be very surprised to hear that people are running their Ember test suites against IE11 (but please let me know if you are!). We currently don't ship an `includes` polyfill for IE11, for example, so if someone's testing it on IE11, they're likely polyfilling a few things anyway.

addon-test-support/dom/find-tooltip-target.js
```diff
@@ -0,0 +1,18 @@
+export function findTooltipTarget(selector) {
+  if (!selector) { // In case of passing null, undefined, etc
+    selector = '.ember-tooltip-target, .ember-popover-target';
+  }
+
+  const { body } = document;
+  const tooltipTarget = body.querySelectorAll(selector);
+
+  if (tooltipTarget.length === 0) {
+    throw Error('ember-tooltips/test-support/dom/find-tooltip-target: No tooltip targets were found.');
```
maxfierke: tiny style thing: I'd like to keep consistent with `throw new Error`, as we use elsewhere

addon-test-support/dom/find-tooltip-target.js
```diff
@@ -0,0 +1,18 @@
...
+  } else if (tooltipTarget.length > 1) {
+    throw Error('ember-tooltips/test-support/dom/find-tooltip-target: Multiple tooltip targets were found. Please provide an {option.targetSelector = ".specific-tooltip-target-class"}');
```
maxfierke: same here

addon-test-support/dom/find-tooltip.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  if (searchMultiple) {
+    return tooltips;
+  }
+
+  let tooltip = tooltips[0];
+
+  if (tooltip && !tooltip.classList.contains('ember-tooltip') && !tooltip.classList.contains('ember-popover')) {
+    throw Error(`getTooltipFromBody(): returned an element that is not a tooltip`);
```
maxfierke: `throw new`

addon-test-support/dom/-get-closest-dom-parent.js
mydea: Yeah, I was a bit unsure about this as well. But if we make new assertion helpers anyway, then this wouldn't be breaking anyhow.

addon-test-support/dom/find-tooltip-target.js
mydea: Ah, that was not on purpose, will fix this!

mydea:
Regarding your general points:

* Good point about the jQuery selectors, I did not think about this. I'll make new assertion helpers then that use the new logic.
* `findTooltip` _does_ now actually always return one element only, unless you pass `{ searchMultiple: true }`. Initially, I have thought of two methods (`findTooltip` and `findTooltips`), but I figured out that `findTooltip()` needs to work with arrays initially anyhow, e.g. when used with the targetSelector stuff, and then the two functions were basically the same, except for the return value, and I went with the option to keep this DRY. 
* Yeah, I can add some tests for the helpers!

addon-test-support/dom/find-tooltip-target.js
mydea: BTW, I basically only copied this over from the existing test helpers, which use `throw Error`. But I do agree that `throw new Error` is nicer!

mydea:
OK, so I've adjusted the PR to:

* Provide new assertion helpers (to be imported from `ember-tooltips/test-support/dom/assertions')
* Add a new test covering these, while leaving the old test for `findTooltip()` intact

addon-test-support/dom/find-tooltip.js
```diff
@@ -0,0 +1,51 @@
+export function findTooltip(selectorOrElement, { targetSelector, searchMultiple = false } = {}) {
```
maxfierke: I think this option should just be shortened to `multiple`

mydea: Changed it!

PR #348, opened 2019-03-14
Make animationDuration configurable

loginovma:
Instead of setting 200 as a magic number, allow to configure
the duration of animation as component property.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ For example, if you want to show a tooltip over a button when the user hovers ov
 
 Options are set as attributes on the tooltip/popover components. Current tooltip/popover properties this addon supports are:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -138,6 +139,18 @@ Options are set as attributes on the tooltip/popover components. Current tooltip
 - [spacing](#spacing)
 - [text (tooltip only)](#text)
 
+#### Animation duration
+
+| Type    | Number  |
+|---------|---------|
+| Default | 200     |
+
+Defines the duration of tooltip animation.
```
maxfierke: We should specify that this is in `ms` and that in testing it will always be `0`.

loginovma: Done. Thanks for the quick response!

PR #350, opened 2019-03-20
Fix issues with inconsistent application of targetSelector option in test helpers

maxfierke:
While writing up docs for the new jQuery-free test helpers, I was upgrading our main app at work, and found a test that started failing with the new helpers, related to the checking for the correct target element when using `targetSelector`. I found that this behavior was actually correct, based on what `targetSelector` is supposed to do, and found that `targetSelector` was being applied inconsistently, resulting in the wrong element being returned in many cases. The jQuery `findTooltip` was not checking the classes of the tooltip/popover target retrieved (in the cases where a selector to the tooltip wrapper is passed as `selector`) with the `targetSelector` specified to the original assertion helper.

This PR updates the jQuery implementation of `findTooltip` to match the semantics of the regular DOM implementation and contains the new docs section on using the new DOM implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Documentation for usage is below:
 - [Actions](#actions)
 - [Testing](#testing)
   - [Test helpers](#test-helpers)
+    - [Importing test helpers](#importing-test-helpers)
 - [Accessibility](#accessibility)
 - [Development](#development)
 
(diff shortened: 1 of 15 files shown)


sir-dunxalot:
I went through the PR  twice now. Nothing stuck out in review.

One thought I had was whether we can simplify the process of finding tooltips with targetSelector to something like `const tooltips = document.querySelectorAll(tooltipSelector);` and then compare those elements' IDs (e.g. `tooltip_uag09h5ywj`) to the target `aria-describedby` attribute of the element returned by `const target = document.querySelector(tooltipSelector)`. However, that seems to just test the DOM structure (e.g. what if the tooltip JS is broken)? And it doesn't seem to have a significant advantage over this approach; nor am I sure if even a it's feasible approach.

From a developer UX point-of-view it seems a pain to change file references. However, I like that you're offering graceful deprecation so that trade off is acceptable IMO.

maxfierke:
@sir-dunxalot yeah, I'm going to take a look at seeing if there's a way to determine whether jQuery integration is enabled at build-time, and have ember-cli dump different helpers into the default path, if that's the case. But I didn't want to hold-up folks who want to be jQuery-free today.

sir-dunxalot:
Maybe it could be a self-declared flag that is added to ember-cli-build.js?

maxfierke:
@sir-dunxalot Looks like [checking jquery-integrration in an addon](https://github.com/emberjs/ember-optional-features#at-build-time-from-an-addon) is documented now. I think this could be paired with `treeForAddonTestSupport` in the `ember-cli-build.js`, and we can output the right thing at build time. I think I'll still merge this, but do a separate PR for that, and update the relevant docs.

sir-dunxalot:
Sounds good!

PR #351, opened 2019-03-29
Build-time determination of test helper type

maxfierke:
* Refactors test helpers to use the same module structure for both
  DOM & jQuery test helpers.
* Uses build-time treeForAddonTestSupport to export either under
  `ember-tooltips/test-support` based on whether jquery-integration
  is enabled in the host application

Allows adoption of DOM test helpers without changing import paths,
provided jQuery is disabled for the host application. Otherwise, either
test helpers are available under their respective `dom` or `jquery`
modules within  `ember-tooltips/test-support`.


package.json
```diff
@@ -29,7 +29,8 @@
     "test:all": "ember try:each"
   },
   "dependencies": {
-    "broccoli-funnel": "^2.0.1",
+    "@ember/optional-features": "^0.6.3",
```
Turbo87: why did you move this to the `dependencies` section?

maxfierke: @turbo87 the optional feature APIs cannot be accessed at build time without it being in dependencies, it seems.

Turbo87: I guess you could assume that if the addon is not available then jQuery wasn't switched off 🤔

maxfierke: Yeah, I think I tried that and it wasn't picking up dependencies from the host app in `this.addons`, but maybe I needed to do something fancy, like traverse my way up to the app or something. I'll dig through the Ember-CLI innards again and see if I can figure out a way to remove the dependency and use the host's, if it's available.

PR #356, opened 2019-05-09
Fix sourcemaps

### LINKED ISSUE #354: Vendor JS sourcemaps missing sourcesContent
Issue #227 appears to be back. Sourcemaps are broken. I have documented the issue with a repo that will reproduce the problem using the latest version of Ember here: https://github.com/les2/soucemap-test-ember-3-9

In short, adding `ember-tooltips` addon causes sourcemaps for vendor to exclude the `sourcesContent` element. This breaks tools such as New Relic Browser, which use `sourcesContent` to unminify JavaScript stacktraces captured in errors (it will not make Network requests to resolve sourcemap, nor is the sourcemap even hosted so that it could).

Relates to https://github.com/sir-dunxalot/ember-tooltips/issues/277

I'm willing to help if you have any pointers.

maxfierke:
popper.js & tooltip.js ship with pre-compiled sourcemaps and we need to include those in order for sourcemap generation to work.

Fixes #354

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -10,6 +10,11 @@ module.exports = function(defaults) {
 
     minifyCSS: {
       enabled: false // CSS minification w/ @import rules seems to be broken in Ember-CLI 3.3
+    },
+
+    sourcemaps: {
+      enabled: true,
+      extensions: ['js']
     }
   });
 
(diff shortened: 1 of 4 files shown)


maxfierke:
This does "fix" it in that it allows vendor.map to be generated, but there seems to be a bunch of missing stuff in there (the contents of `addon-tree-output`). Looking into that that.

maxfierke:
Fixed in 87984d0

PR #384, opened 2019-10-23
Replace ember-wormhole with ember-maybe-in-element

maxfierke:
This is a lighter weight dependency and works using the `in-element` helper built into Glimmer itself.

Supersedes #319

### DIFF
--- a/addon/components/ember-tooltip-base.js
+++ b/addon/components/ember-tooltip-base.js
@@ -181,8 +181,22 @@ export default Component.extend({
 
   /* An ID used to identify this tooltip from other tooltips */
 
-  wormholeId: computed('elementId', function() {
-    return `${this.get('elementId')}-wormhole`;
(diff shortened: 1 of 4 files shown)


Turbo87:
unfortunately this seems to be a breaking change for us since we use ember-basic-dropdown in our app and that uses a different version of `ember-maybe-in-element`. now ember-cli-dependency-lint is complaining... 😞

maxfierke:
@Turbo87 looks like the latest ember-basic-dropdown is using the same version, if you're able to upgrade.

I don't believe there's much (if any) of an API difference between versions of `ember-maybe-in-element`, so you might just want to add an [`allowedVersions`](https://github.com/salsify/ember-cli-dependency-lint#allowed-versions) configuration for `ember-maybe-in-element` to allow both the version ember-tooltips uses and the one the version that ember-basic-dropdown uses

PR #385, opened 2019-12-17
Remove for..of loop for IE11 / no Symbol support

raido:
`for...of` loop requires `Symbol` to exist which means it requires polyfills for older browsers like IE11 and others.

With changes in this PR `Symbol` existence is no longer a requirement.


raido:
🤞for v3.4.1 patch release.

maxfierke:
I'll tag a release later tonight, thanks!

Issue #310, opened 2018-11-05
IE11 rendering issues (Includes & Symbol)


transitorytammy:
Right now the `ember-tooltips` doesn't render on IE 11 due to `includes` not being compatible with with IE11.  

https://caniuse.com/#search=includes

I think a solution would be to switch to the use of `indexOf`.  Also, the other issue I found in IE was that `Symbol` is undefined.  Adding the babel polyfill would help with that issue too.  https://github.com/babel/ember-cli-babel/issues/40#issuecomment-268952820

 I would be happy to make a PR to make this change if you in agreement with these changes.  :)

maxfierke:
Hmm... it doesn't look like we directly use `Symbol` anywhere in ember-tooltips. Do you know where that use is coming from?

I'd be open to switching the use of `includes` over to `indexOf`, since it's so limited, if it would make it easier to use `ember-tooltips` in IE11 without the inclusion of polyfills.

However, w/r/t inclusion of the babel-polyfill, I think that decision should be differed to the apps using `ember-tooltips`, as it's fairly hefty. If the polyfill is required in order to cover `Symbol` then that should maybe be our guidance for `includes` as well.

maxfierke:
I'm going to go ahead and close this out, but feel free to re-open if there are other questions. I think the guidance on leveraging `babel-present-env` in the host application to include the appropriate polyfills still stands.

PR #396, opened 2020-05-13
Array.includes -> Array.indexOf for IE11

raido:
Addresses: https://github.com/sir-dunxalot/ember-tooltips/issues/310

Symbol issue from for..of loop mention in that ticket has been solved in previous releases already.


raido:
Thank you!

PR #415, opened 2020-12-05
Accessibility: Do not hide popover when focusing an element within the popover

johanrd:
Fixes an accessibility issue where the popover hides if an element within the popover is navigated to by keyboard.

```hbs
<button>Tab into this button, then press Enter to open popover
  <EmberPopover @event="click">
    <button>Click Tab to move focus to this button</button> <!-- the popover should not hide -->
    <button>Click Tab again to move focus to this button</button> <!-- the popover should not hide -->
  </EmberPopover>
</button>
```

The fix checks if the relatedTarget of the 'focusout' event is inside inside the popover. If true, the popover is not hidden.

### DIFF
--- a/addon/components/ember-popover.js
+++ b/addon/components/ember-popover.js
@@ -57,8 +57,8 @@ export default EmberTooltipBase.extend({
       this.set('_isMouseInside', false);
     });
 
-    this._addEventListener('focusout', () => {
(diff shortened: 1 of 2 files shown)


maxfierke (commented):
looks good to me! Mind adding some quick tests for both tooltips and popovers to protect from regressions?

PR #418, opened 2021-02-08
chore: update ember-in-element-polyfill

Techn1x:
There was a recent release of ember-in-element-polyfill which fixes usage in monorepos / yarn workspaces.

(based on the new yarn.lock, looks like it also had dependabot/maintenance bumps of a few packages too)


maxfierke:
Looks good to me, but I'm not sure this is wholly required for consuming applications, as 1.0.1 satisfies the current `^1.0.0` constraint. May just need to use `1.0.1` in the parent project and/or remove the pinned `1.0.0` resolution in your current `yarn.lock` and re-run `yarn` (or use something like [`yarn-deduplicate`](https://github.com/atlassian/yarn-deduplicate))

I'll still get this in, but I need to do some cleanup work to get Ember-CLI updated and CI moved over to GitHub actions first, so might be later in the week.

Techn1x:
You're totally right - no rush on this PR. Thanks for looking at this!

maxfierke:
@Techn1x master is updated w/ GitHub Actions for CI (as well as the latest Ember-CLI). Mind giving this a rebase and then I'll merge (assuming CI passes)?

Techn1x:
ooo nice work.
PR updated. Commits are a little messy, are you able to squash when you merge this with the merge button? otherwise I can squash it

maxfierke (approved):

PR #420, opened 2021-02-26
Use "this.import" instead of "app.import" to support lazy engines

### LINKED ISSUE #363: ember-tooltips 3.3.1+ build breaks w/ lazy engines
Upgrading our application from 3.3.0 -> 3.3.2, and we're getting build failures:
```
  - broccoliBuilderErrorStack: Error: ENOENT: no such file or directory, lstat '/var/folders/89/kt8fhhgx0pz0m3s5s588tkvc0000gn/T/broccoli-42794yLQNVhp1CiNT/out-016-broccoli_merge_trees_empty_merge_tree/vendor/ember-tooltips--popper.js'
```

Unfortunately I do not have time to dig into this more right now, but we do have an app with several lazy-load engines, and at least one does include `ember-tooltips`, so I suspect this may be the issue.

bertdeblock:
Fixes #363.

I'm no Ember CLI expert but this fixes the lazy engines issue for me.
Based on [this ember-cli-moment-shim commit](https://github.com/jasonmit/ember-cli-moment-shim/commit/6f3dea3c7db8d6220e8dff393c4975139ada4611).

I've also tested the following scenario's locally to ensure the imports still work properly:
- `ember-tooltips` as a dev and normal dependency of an addon
- `ember-tooltips` as a dev and normal dependency of a project
- `ember-tooltips` as a nested dependency of a project

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,10 +29,10 @@ module.exports = {
     }
   },
 
-  included: function (app) {
+  included: function () {
     this._super.included.apply(this, arguments);
 
-    app.import('vendor/ember-tooltips--popper.js', {
+    this.import(`${this.treePaths.vendor}/ember-tooltips--popper.js`, {
       using: [
         {
           transformation: 'amd',
@@ -40,7 +40,7 @@ module.exports = {
         },
       ],
     });
-    app.import('vendor/ember-tooltips--tooltip.js', {
+    this.import(`${this.treePaths.vendor}/ember-tooltips--tooltip.js`, {
       using: [
         {
           transformation: 'amd',


maxfierke (commented):
nice, I'll give it a spin again against a reproduction I have.

As an aside: I remember trying `this.import` as a solution to this, but I think there may have been some gotcha with it, either something to do with Yarn Workspaces or maybe a bug within ember-engines that has since been fixed

bertdeblock:
@maxfierke Thanks, that would be great! The project I'm working on also uses Yarn Workspaces. Do you know why the test suite fails atm? I'll try to have a look at it this weekend or monday.

maxfierke:
> @maxfierke Thanks, that would be great! The project I'm working on also uses Yarn Workspaces. Do you know why the test suite fails atm? I'll try to have a look at it this weekend or monday.

nope, no idea. Initial looks like some sort of dependency updated underneath, maybe unanchored by ember-try. Was working last time I ran the suite a few weeks ago, so would assume it'll might get resolved on its own but I'll try to take a closer look (probably same timeframe)

bertdeblock:
Downgrading `ember-in-element-polyfill` to v1.0.0 seems to fix it, so that should already narrow it down a bit.

maxfierke:
I might just remove 2.18 from the CI matrix (maybe just temporarily). At this point it's over three years old, and I'm not really interested in figuring out why sometime in the last 20 days, it's suddenly failing CI, so I'll get that updated in master. Then you should be able to rebase and see some green checks.

maxfierke:
Removed the 2.18 scenario in #421

maxfierke:
Looking at that the ember-cli-moment-shim example, it occurs to me that maybe we should be using `this.treePaths.vendor` instead of `vendor` in the import path too 🤔

bertdeblock:
@maxfierke Updated accordingly.

maxfierke (approved):
:octocat: looks good. Will merge and release later today

bertdeblock:
Thank you! 👍

PR #429, opened 2021-05-26
Add @embroider/test-setup to ember-try scenarios

thoov:
Adds two new scenarios for ember-try: embroider-safe and embroider-optimized. Embroider-safe does not enable any tree shaking optimizations where as embroider-optimized enables them all.

ember-cli/ember-cli#9536


maxfierke (approved):
:octocat: awesome, thanks!

PR #437, opened 2021-12-25
Remove deprecated assign polyfill

mnutt:
Ember 4.x now recommends using `Object.assign`: https://deprecations.emberjs.com/v4.x/#toc_ember-polyfills-deprecate-assign


maxfierke:
Hi @mnutt, thanks for the contribution! Unfortunately, this would break IE 11 compatibility, so we will need to wait until a major release to merge this. The next major release is probably going to change a bunch of stuff about the underlying implementation anyway and may remove the need for the popper options merging code.

In the meantime, if there's a way to avoid the deprecation while retaining compatibility for IE 11, I'm open to it.

PR #439, opened 2022-01-30
Remove `assign()` usage

Turbo87:
The `assign()` function in Ember.js has been deprecated with v4. This replaces it with the object spread syntax which should be compiled away by Babel, if necessary.

This is an alternative to and closes https://github.com/sir-dunxalot/ember-tooltips/pull/437

/cc @maxfierke


maxfierke (approved):
:octocat: lgtm! I'm assuming the transform is built into any relatively modern ember-cli-babel w/ in the version constraint, but let me know if you know otherwise

Turbo87:
> the transform is built into any relatively modern ember-cli-babel w/ in the version constraint, but let me know if you know otherwise

yep, that is my understanding too

PR #446, opened 2022-06-24
Add `targetElement` argument

bendemboski:
Add a `targetElement` argument that's similar to `targetId`, but can specify an HTML element for cases where the target element doesn't have an ID, e.g. HTML rendered by a third-party library.

My use case for this is attaching a popover to the currently-selected paragraph in content managed by a CKEditor instance to present options for acting on the paragraph, e.g. block formatting such as alignment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,15 @@ The tooltip will always be rendered on its parent element unless you specify the
 </EmberTooltip>
 ```
 
+or the `targetElement` attribute:
+
+```hbs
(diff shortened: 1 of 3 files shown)


maxfierke (commented):
@bendemboski This looks great! CI should be fixed if you rebase off master (looks like Node 12 fell out of "maintained node versions" in browserlist)

maxfierke (approved):
:octocat: !

bendemboski:
Sweet, thanks for the quick approval!

PR #462, opened 2025-02-02
fix: Error: Could not find module '@ember/string' on ember 6.1.0

amiarSlimane:
### Description

This pull request addresses an issue where the module `@ember/string` could not be found in Ember 6.1.0.

### Changes

- **package.json**
  - Updated `ember-auto-import` from `^1.11.3` to `^2.10.0`
  - Added `@ember/string` with version `^3.1.1`

- **ember-try.js**
  - Updated `fastboot` resolution from `^3.1.2` to `^4.1.5` in both `embroiderSafe` and `embroiderOptimized` configurations

### Testing

- Ensure that the application builds successfully with the updated dependencies.
- Verify that the module `@ember/string` is correctly resolved and no longer causes errors in Ember 6.1.0.
- Run existing test suites to verify that there are no regressions (versions prior to 2.24 are failing).

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -107,14 +107,14 @@ module.exports = async function () {
       embroiderSafe({
         npm: {
           resolutions: {
-            '**/fastboot': '^3.1.2',
+            '**/fastboot': '^4.1.5',
           },
         },
       }),
       embroiderOptimized({
         npm: {
           resolutions: {
-            '**/fastboot': '^3.1.2',
+            '**/fastboot': '^4.1.5',
           },
         },
       }),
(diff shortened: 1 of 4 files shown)


mansona (approved):
For what it's worth this looks like the right kind of change 👍 If we could get an approval to run CI that would be great, if it passes it's probably worth merging and releasing

maxfierke:
Sorry for the delay, kicked off a CI run -- I'm fairly out of the Ember world at this point, but happy to usher in a new release to keep things working

amiarSlimane:
I have started migrating this addon to V2 and i found that tooltip.js is deprecated [https://www.npmjs.com/package/tooltip.js](https://www.npmjs.com/package/tooltip.js) , is it a good idea to do a migration for it ?

mansona:
@amiarSlimane its probably worth doing one and then the other, rather than doing both at the same time. i.e. v2 addon still with tooltip.js and then another later release migrating 👍

PR #465, opened 2025-06-09
Modernize dummy app to use native classes

raido:
This PR modernizes dummy application to use native classes.


tests/dummy/app/components/some-component.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="some-component">
```
raido, line 1: Turned into template only component.

tests/dummy/app/controllers/index.js
```diff
@@ -1,40 +1,48 @@
...
+      }, 2000);
+    });
+  }
+
+  @action
+  togglePopover() {
+    this.showToggleablePopover = !this.showToggleablePopover;
+  }
+
+  scheduleShowLogoTooltip = modifier(() => {
```
raido, line 36: Added ember-modifier to dev dependencies and implemented as modifier instead.

tests/dummy/app/controllers/update-for-reposition.js
```diff
@@ -1,43 +1,44 @@
...
 
-    deleteData() {
-      set(this, 'isLoading', true);
-      set(this, 'rows', []);
-    },
-  },
-});
+  @action
+  deleteData() {
+    if (!this.isLoading) {
```
raido, line 39: Slight modification as in tracked world without this guard we'd end up with infinite loop for updateFor attribute.

tests/dummy/app/routes/application.js
```diff
@@ -1,20 +0,0 @@
-import Route from '@ember/routing/route';
-import { capitalize } from '@ember/string';
-
-export default Route.extend({
```
raido, line 4: Seemed to be unused, I could not find any {{action}} or send() calls with those actions.

